PR #14, opened 2021-09-28
Switch to numba .54

Hardcode84:


Hardcode84:
@AlexanderKalistratov @geexie @Vyacheslav-Smirnov FYI

PR #17, opened 2021-10-02
Update to MLIR main

Hardcode84:


Hardcode84:
@AlexanderKalistratov @geexie @Vyacheslav-Smirnov FYI

PR #46, opened 2021-11-25
Fix mem corruption in SCFToAffine

Hardcode84:
Temporary object returned by `getMultiDimIdentityMap` was already dead by the time `makeArrayRef` returned.


Hardcode84:
@geexie FYI

PR #113, opened 2022-01-19
Move numba_dpcomp to subdir

Hardcode84:
* Move frontend code into subdirectory to remove python-specific files from top-level dir.


Hardcode84:
@diptorupd @Jianhui-Li FYI

diptorupd:
@Hardcode84 please hold off on merging till others have had a chance to review and comment. As more people are going to be involved can we enforce PRs have a proper description to make it easier to review?

Hardcode84:
@diptorupd  Only Jianhui-Li  sent me github id, so others probably not much interested yet. Anyway, keeping these rename PRs alive for long is too disruptive. I will accept post review and can do changes in follow up PRs.

PR #211, opened 2022-04-27
[Do not merge] Initial support for auto updating and building dependent third party projects

silee2:


diptorupd:
@silee2 Please add a description, I want to understand what you are attempting. I too am making some changes to build system, so we should make sure our changes do not collide.

 Since you had a `Do not merge` in the title, I converted it to draft.

diptorupd:
@silee I am unable to leave any review comments, maybe my access is not setup correctly.

CMakeLists.txt
```diff
@@ -70,6 +72,43 @@ macro(apply_llvm_compile_flags target)
...
+    else()
+        set(MLIR_EXT_USE_EXT_LEVEL_ZERO OFF)
+    endif()
+    message(STATUS "MLIR_EXT_USE_EXT_LEVEL_ZERO: ${MLIR_EXT_USE_EXT_LEVEL_ZERO}")
+endif()
+
+add_subdirectory(third-party)
+
+if(NOT MLIR_EXT_USE_LLVM_INSTALL_TREE)
+    set(LLVM_DIR ${PROJECT_BINARY_DIR}/third-party/mlir/llvm-project/llvm/lib/cmake/llvm)
```
diptorupd, line 101: I do not like checking out third party repos in IMEX repo. The problem is if you use `git clean -dfx` the folder will be wiped out. In my Python build_locally.py, I allow users to set a `WORKING_DIR` or use system temp directory. Let us at least have the option of providing a -DCMAKE_THIRD_PARTY_LIB_PATH` or something like that?

CMakeLists.txt
```diff
@@ -70,6 +72,43 @@ macro(apply_llvm_compile_flags target)
...
+elseif (DEFINED LLVM_DIR)
+    message(FATAL_ERROR "Need to define both LLVM_DIR and MLIR_DIR or none.")
+else()
+    set(MLIR_EXT_USE_LLVM_INSTALL_TREE OFF)
+endif()
+
+if(GPU_ENABLE)
+    # Check if using external level-zero
+    if(DEFINED LEVEL_ZERO_DIR)
+        set(MLIR_EXT_USE_EXT_LEVEL_ZERO ON)
```
diptorupd, line 91: Let us move to `IMEX` as the prefix of all our CMake flags.

level-zero-sha.txt
```diff
@@ -0,0 +1 @@
+bb7fff05b801e26c3d7858e03e509d1089914d59
```
diptorupd, line 1: Perhaps bind to a released tag for L0.

CMakeLists.txt
```diff
@@ -70,6 +72,43 @@ macro(apply_llvm_compile_flags target)
     target_compile_definitions(${target} PRIVATE ${LLVM_DEFINITIONS})
 endmacro()
 
+# Check if using external LLVM/MLIR install tree
+if (DEFINED MLIR_DIR)
```
diptorupd, line 76: We should really not need two flags. If someone is using an LLVM build without building MLIR project, CMake will throw an error and we should at that point inform user what to do.

diptorupd: Perhaps I am not understanding it right. What does `MLIR_DIR` point to?

diptorupd (commented):
I had considered using ExternalProject too. What I did not like was that I needed to first to a `make llvm` and then `make imex`. 

But, I like the overall direction of the changes. They will not break my Python build script and direct usage of CMake will work as well.

CMakeLists.txt
silee2: MLIR_DIR is defined to guide "find_package(MLIR)" work correctly in CMake.
LLVM_DIR is for "find_package(LLVM)"
Both were already used in the code base. Not introduced by this PR.
Actually find_package(MLIR) will call find_package(LLVM) internally. So we can clean it up but that requires more changes.

CMakeLists.txt
silee2: OK

CMakeLists.txt
silee2: The PR does not remove the option of providing -DCMAKE_THIRD_PARTY_LIB_PATH. It only checks out third party repo if -DCMAKE_THIRD_PARTY_LIB_PATH is not provided.

PR #224, opened 2022-06-02
Add flag for reference counting

nbpatel:
Reference counting is required by numba and hence adding it in the IMEX_ENABLE_NUMBA_HOTFIX flag

### DIFF
--- a/dpcomp_gpu_runtime/lib/gpu_runtime.cpp
+++ b/dpcomp_gpu_runtime/lib/gpu_runtime.cpp
@@ -346,7 +346,9 @@ struct Stream {
       throw std::runtime_error("Failed to allocate MemInfo");
     }
 
-    retain();
+    #if defined(IMEX_ENABLE_NUMBA_HOTFIX)
+       retain();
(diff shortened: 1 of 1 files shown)


dpcomp_gpu_runtime/lib/gpu_runtime.cpp
```diff
@@ -346,7 +346,9 @@ struct Stream {
       throw std::runtime_error("Failed to allocate MemInfo");
     }
 
-    retain();
+    #if defined(IMEX_ENABLE_NUMBA_HOTFIX)
```
diptorupd, line 349: @nbpatel why do we even have numba_dpcomp-specific things in `gpu_runtime`? I am fine with the quick fix, but we need a plan to refactor/redesign the code.

nbpatel: I guess it is there for the python frontend. Agree, we need to refactor it

diptorupd (approved):

PR #234, opened 2022-06-21
Fschlimb/addconversion

### LINKED ISSUE #233: IMEX needs a CLI tool similar to mlir-opt
The `mlir-opt` tool makes it easy to run any MLIR pass using a default global registry. The tool is useful to design `FileCheck` test cases.

IMEX needs a similar tool. The following pieces need to be put in place before the tool can be put together:

- [x] Add a `Passes.td` and `Passes.h` to our `Conversion` folder. These two files are provided by the MLIR repo and are used to define the registration functions for all `Conversion` passes.
- [x] Add `Passes.td` and `Passes.h` to all IMEX `Dialect` that define passes.
- [x] Add an `InitIMEXDialects.h` header similar to the `InitAllDialects.h` in MLIR.
- [x] Add an `InitIMEXPasses.h` header similar to `InitAllPasses.h` in MLIR.
- [x] Add an `imex-opt.cpp` source similar to `mlir-opt.cpp`

Explore automation possibilities to auto-generate these files as most of the files will need mechanical maintenance.

PS: The automation scripts in our repo are `meta^2` programming. Use Python to generate TableGen that then generates C++. Ergo, Python rules them all!!

fschlimb:
Basics for aligning with MLIR conventions in pass infrastructure.
- Using tablegen for Pass' base classes
- Header-files bringing passes together
- Moving all imex dialects and passes into namespace imex
- adjusting ad_*.py scripts

Closes #233


fschlimb:
#233 mostly done (the current stub has no in-dialect passes so that part is missing).

Before merging, I suggest we use imex-opt on a few simple tests.

diptorupd:
@fschlimb \o/ \o/ thank you very much. I will try it out and also review the PR tomorrow.

fschlimb:
add_dialect now also generates stubs and cmakes for Dialect passes/transforms. Instructions updated in README.

When we add the first dialect with the scripts for real, we might find some issues in the template code and should it fix the scripts as we go.

diptorupd:
All done. Merging. Thanks @fschlimb for getting all of this done.

PR #266, opened 2022-07-21
POC for adding a sycl runtime and Intel Gpu dialect

nbpatel:

This PR adds a sycl runtime and also a new dialect called Intel Gpu dialect. We do not use the gpu_runtime dialect and directly go from GPU dialect to llvm runtime calls. 
I have tested it locally with the test cases we have and it works.


dpcomp_gpu_runtime/lib/gpu_runtime.cpp
```diff
@@ -19,6 +19,7 @@
 #include <cstdint>
 #include <cstdio>
 #include <cstdlib>
+#include <iostream>
```
Hardcode84, line 22: Including iostream comes with compile time and binary size costs. It is fine for examples but please do not use in runtime or compiler.

Hardcode84:
Some general comments:
* I would prefer to have both L0 and CYCL gpu runtimes intact for some time, compiled in separate shared libs. Also, in this case you don't need separate runner, you can just pass different runtime lib to existing runner.
* Gpu dialect reworking should be separated in its own PR

Hardcode84:
* Let's keep l0 and SYCL runtime in separate shared libs.
* Let`s refactor `gpu_runtime` dialect into `igpu` in separate PR. And any differences in lowering should be done via pass options instead of copypasting entire passes.
* I would prefer for SYCL runtime to have same API as L0 one. It will allow to switch api by simple switching shared library and will eliminate need for separate runner.

Hardcode84:
Let's do this in multiple steps:
* Raname `gpu_runtime` dialect into `intel_gpu` (without functional changes) in separate PR as we definitely do not want to have 2 dialects with 90% copypaste
*  For any lowering differences between L0 and SYCL make a pass option instead of copypasting entire passes

Hardcode84:
Initial idea for SYCL support was to just replace L0 `gpu_runtime` with SYCL `gpu_runtime` without changing lowering pipeline, and I wasn't provided with clear justification why this is not possible and why we went current route instead.

Drastic changes to GPU lowering pipeline will introduce additional support burden for our Python pipeline which I'm not ready to take.

nbpatel:
closing this draft. adding changes incrementally to the refactor branch

PR #269, opened 2022-07-26
Python pipeline refactoring and cleanup

Hardcode84:
* Do not build local FileCheck if external is present (it will fail due to target name conflict)
* Extract `LowerArgOps` pattern from `PlierToScfPass` into separate pass, `PlierToScfPass` no longer uses any python ops (it is still uses `UndefOp` though)
* Separate utils ops llvm lowering  from python specific lowering so it can be later used outside python pipeline (required by gpu pipeline, will be done in separate PR)

### DIFF
--- a/numba_dpcomp/numba_dpcomp/mlir_compiler/lib/pipelines/lower_to_llvm.cpp
+++ b/numba_dpcomp/numba_dpcomp/mlir_compiler/lib/pipelines/lower_to_llvm.cpp
@@ -1755,6 +1755,7 @@ struct LowerReleaseContextOp
   }
 };
 
+// Convert operations from the plier_util dialect to the LLVM dialect.
 struct PierUtilToLLVMPass
     : public mlir::PassWrapper<PierUtilToLLVMPass,
                                mlir::OperationPass<mlir::ModuleOp>> {
--- a/numba_dpcomp/numba_dpcomp/mlir_compiler/lib/pipelines/lower_to_llvm.cpp
+++ b/numba_dpcomp/numba_dpcomp/mlir_compiler/lib/pipelines/lower_to_llvm.cpp
@@ -1787,7 +1787,7 @@ struct PierUtilToLLVMPass
 
     mlir::LLVMConversionTarget target(context);
     target.addLegalOp<mlir::func::FuncOp>();
-    target.addIllegalOp<plier::TakeContextOp, plier::ReleaseContextOp>();
+    target.addIllegalDialect<plier::PlierUtilDialect>();
     if (failed(applyPartialConversion(op, target, std::move(patterns))))
       signalPassFailure();
   }
--- a/numba_dpcomp/numba_dpcomp/mlir_compiler/lib/pipelines/lower_to_llvm.cpp
+++ b/numba_dpcomp/numba_dpcomp/mlir_compiler/lib/pipelines/lower_to_llvm.cpp
@@ -1755,6 +1755,43 @@ struct LowerReleaseContextOp
   }
 };
 
+struct PierUtilToLLVMPass
+    : public mlir::PassWrapper<PierUtilToLLVMPass,
+                               mlir::OperationPass<mlir::ModuleOp>> {
+  MLIR_DEFINE_EXPLICIT_INTERNAL_INLINE_TYPE_ID(PierUtilToLLVMPass)
+
+  PierUtilToLLVMPass() = default;
+
+  void runOnOperation() override {
+    mlir::Operation *op = getOperation();
+    auto &context = getContext();
+    auto options = getLLVMOptions(context);
+
+    mlir::LLVMTypeConverter typeConverter(&context, options);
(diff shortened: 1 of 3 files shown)


numba_dpcomp/numba_dpcomp/mlir_compiler/lib/pipelines/lower_to_llvm.cpp
```diff
@@ -1755,6 +1755,43 @@ struct LowerReleaseContextOp
   }
 };
 
+struct PierUtilToLLVMPass
```
fschlimb, line 1758: Please add docs - at least to new entities - by adding doxygen comments describing high-level purpose and functionality as well as args/return types.

fschlimb (changes requested):

numba_dpcomp/numba_dpcomp/mlir_compiler/lib/pipelines/lower_to_llvm.cpp
Hardcode84: done

numba_dpcomp/numba_dpcomp/mlir_compiler/lib/pipelines/plier_to_scf.cpp
```diff
@@ -595,6 +595,9 @@ struct CondBranchSameTargetRewrite
   }
 };
 
+// Convert plier::ArgOp into direct function argument access. ArgOp is just an
+// artifact of Numba IR conversion and doens't really have any functional
+// meaning so we can get rid of ot early.
```
fschlimb, line 600: get rid of ot -> git rid of it

numba_dpcomp/numba_dpcomp/mlir_compiler/lib/pipelines/lower_to_llvm.cpp
```diff
@@ -1755,6 +1755,44 @@ struct LowerReleaseContextOp
...
+        ExpandShapeLowering,
+        LowerExtractMemrefMetadataOp,
+        LowerTakeContextOp,
+        LowerReleaseContextOp
+        // clang-format on
+        >(typeConverter);
+
+    mlir::LLVMConversionTarget target(context);
+    target.addLegalOp<mlir::func::FuncOp>();
+    target.addIllegalOp<plier::TakeContextOp, plier::ReleaseContextOp>();
```
fschlimb, line 1790: If this is lowering plier_util to llvm, why doesn't it declare plier_util as illegal?

Hardcode84: This was mostly covered by input/output types legality check already present `LLVMTypeConverter`, but it won't hurt I guess.

numba_dpcomp/numba_dpcomp/mlir_compiler/lib/pipelines/plier_to_scf.cpp
Hardcode84: Done

numba_dpcomp/numba_dpcomp/mlir_compiler/lib/pipelines/lower_to_llvm.cpp
Hardcode84: done

numba_dpcomp/numba_dpcomp/mlir_compiler/lib/pipelines/plier_to_scf.cpp
```diff
@@ -630,7 +636,7 @@ void PlierToScfPass::runOnOperation() {
   auto op = getOperation();
   (void)mlir::applyPatternsAndFoldGreedily(op, std::move(patterns));
 
-  op.walk([&](mlir::Operation *o) -> mlir::WalkResult {
+  op->walk([&](mlir::Operation *o) -> mlir::WalkResult {
```
fschlimb, line 639: Is this the same as a check for legal ops?

fschlimb (approved):

numba_dpcomp/numba_dpcomp/mlir_compiler/lib/pipelines/plier_to_scf.cpp
Hardcode84: We cannot use legality check here as we don't  use dialect conversion framework.

fschlimb: I am just trying to understand.
Shouldn't this use the conversion framework?

Hardcode84: This is actually much easier without conversion framework because you don't need to write so much legality checks

PR #276, opened 2022-07-29
Fix Python gpu tests

Hardcode84:
* `LowerTakeContextOp`: insert proper type conversion on function call boundaries
* `ConvertAtomicOps`: increase pattern benefit. We are representing gpu atomic ops as calls to functions with some magical names, but upstream recently added theirs own lowering for functions calls in spirv. Increase benefit of our pattern so it is always running first.

### DIFF
--- a/mlir/lib/Conversion/gpu_to_gpu_runtime.cpp
+++ b/mlir/lib/Conversion/gpu_to_gpu_runtime.cpp
@@ -829,7 +829,10 @@ static mlir::Value lowerFloatSubAtomic(mlir::OpBuilder &builder,
 
 class ConvertAtomicOps : public mlir::OpConversionPattern<mlir::func::CallOp> {
 public:
-  using OpConversionPattern::OpConversionPattern;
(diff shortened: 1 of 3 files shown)


fschlimb:
I am happy to review this but I need a few code-comments to understand the flow and intent.

Hardcode84:
I've added some comments but proper documentation is out of scope of this PR.
And I'll merge this later today anyway as without it our end-to-end python gpu pipeline is broken.

fschlimb:
> And I'll merge this later today anyway as without it our end-to-end python gpu pipeline is broken.

Why are you asking for a review then?

mlir/lib/Conversion/gpu_to_gpu_runtime.cpp
```diff
@@ -829,7 +829,10 @@ static mlir::Value lowerFloatSubAtomic(mlir::OpBuilder &builder,
 
 class ConvertAtomicOps : public mlir::OpConversionPattern<mlir::func::CallOp> {
```
fschlimb, line 830: convert to what?

PR #277, opened 2022-08-02
Update llvm sha

Hardcode84:
Needed to pull some fixes from upstream.
Changes included
* `InvertCmpi` pattern was upstreamed (https://reviews.llvm.org/D129929)
* Type was removed from mlir `Attribute` class, cast to `TypedAttr` where needed
* `spv.OCL` ops were named to `spv.CL`
* Memref reshape llvm lowering was fixed upstream so our pattern no longer needed (https://reviews.llvm.org/D129965)
* Invalid tensor cast folding was fixed in upstream (https://reviews.llvm.org/D130314), reenable affected tests, one test still failing due to issue in `PropagateLayout` pass, will be addressed in separate PR


fschlimb (approved):

nbpatel (approved):
LGTM. Ivan said the Gpu e2e tests pass.

Hardcode84:
There was an issue with zstd (see failed tests), and it seems it is a misconfiguration on conda-forge side and a known issue (https://github.com/conda-forge/zstd-feedstock/issues/58). Disabling zstd in llvm helped and shouldn't affect us.

fschlimb:
@Hardcode84 Do you know why the llvm-cache name has "1" as the cache number? The build.yml says "0", not "1".

fschlimb (changes requested):
@Hardcode84 The build.yml still has an issue with the cache names. Apparently it re-uses the env var CACHE_NUMBER from an earlier step within the build of LLVM. To fix this, we need separate CACHE_NUMBER vars for each step, e.g. change build.yml lines 85 and 89 to use LLVM_CACHE_NUMBER and set it to 1 in line 85 to avoid re-building llvm.

fschlimb (approved):
Thanks @Hardcode84 !

PR #279, opened 2022-08-03
update refactor to latest llvm

fschlimb:
update to latest llvm (following main #277 )


Hardcode84:
Why build got cancelled?

fschlimb:
> Why build got cancelled?

The cache policy [does not seem to allow re-using cache from branches which are not parents (or main)](https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#restrictions-for-accessing-a-cache). I'll have to wait until the [build on main](https://github.com/intel/mlir-extensions/runs/7650937438?check_suite_focus=true) completes. It also does not re-use the cache created in the PR! Pretty limited feature...

Hardcode84 (approved):

nbpatel (approved):

PR #282, opened 2022-08-04
Option e2e test

lchang552:

Add an environment variable for turning on and off the end-to-end tests of the core compiler. Users can turn it on by using 'export IMEX_ENABLE_GPU_E2E_TEST=0' or turn it off using 'export IMEX_ENABLE_GPU_E2E_TEST=1' according to if targe GPU exists or not.

### DIFF
--- a/mlir/test/L0_RUNNER/lit.local.cfg
+++ b/mlir/test/L0_RUNNER/lit.local.cfg
@@ -1,1 +1,2 @@
-config.unsupported = True
+if not config.enable_e2e_gpu_test:
+    config.unsupported = True
--- a/mlir/test/L0_RUNNER/lit.local.cfg
+++ b/mlir/test/L0_RUNNER/lit.local.cfg
@@ -1,3 +1,6 @@
-if not config.enable_e2e_gpu_test:
+import os
+
+if os.environ.get('IMEX_ENABLE_GPU_E2E_TEST') == "1":
+    config.unsupported = False
+else:
     config.unsupported = True
-    
--- a/mlir/test/L0_RUNNER/lit.local.cfg
+++ b/mlir/test/L0_RUNNER/lit.local.cfg
@@ -1,2 +1,3 @@
 if not config.enable_e2e_gpu_test:
     config.unsupported = True
+    
--- a/mlir/test/L0_RUNNER/lit.local.cfg
+++ b/mlir/test/L0_RUNNER/lit.local.cfg
@@ -1,6 +1,3 @@
 import os
 
-if os.environ.get('IMEX_ENABLE_GPU_E2E_TEST') == "1":
-    config.unsupported = False
-else:
-    config.unsupported = True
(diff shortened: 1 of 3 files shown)


Hardcode84 (changes requested):
I'd prefer it to be an env var checked by lit itself to enable/disable tests without reconfiguring project

lchang552:
> I'd prefer it to be an env var checked by lit itself to enable/disable tests without reconfiguring project

This way of configuration-based variable passing is inspired from upstream here: https://github.com/llvm/llvm-project/blob/main/mlir/test/Integration/lit.local.cfg

I also think checking the env var is nice so that we can automatically configure it. But I'm not sure where to find the provided env var so we can check it. Should I refer to the documents of the level zero driver?

Hardcode84:
> I also think checking the env var is nice so that we can automatically configure it. But I'm not sure where to find the provided env var so we can check it. Should I refer to the documents of the level zero driver?

I mean something like
```
export IMEX_ENABLE_GPU_E2E_TEST=1
ninja check-l0-runner
```

And lit config is just a usual python script, you can check env vars there using usual python `os.environ.get('IMEX_ENABLE_GPU_E2E_TEST')`

lchang552:
> > I also think checking the env var is nice so that we can automatically configure it. But I'm not sure where to find the provided env var so we can check it. Should I refer to the documents of the level zero driver?
> 
> I mean something like
> 
> ```
> export IMEX_ENABLE_GPU_E2E_TEST=1
> ninja check-l0-runner
> ```
> 
> And lit config is just a usual python script, you can check env vars there using usual python `os.environ.get('IMEX_ENABLE_GPU_E2E_TEST')`

Thanks! This makes sense to me. I will do it in this way.

mlir/test/L0_RUNNER/lit.local.cfg
```diff
@@ -1 +1,6 @@
-config.unsupported = True
\ No newline at end of file
+import os
+
+if os.environ.get('IMEX_ENABLE_GPU_E2E_TEST') == "1":
+    config.unsupported = False
+else:
+    config.unsupported = True
```
Hardcode84, line 6: `config.unsupported = bool(int(os.environ.get('IMEX_ENABLE_GPU_E2E_TEST', 0)))`

Hardcode84: Also, rename it to 'IMEX_ENABLE_GPU_E2E_TESTS'

Hardcode84 (approved):

mlir/test/L0_RUNNER/lit.local.cfg
lchang552: OK, done

Hardcode84: I forgot `not` :)

lchang552: Haha, I see

Hardcode84 (approved):

nbpatel (approved):

PR #283, opened 2022-08-05
initial PTensor RFC - draft

fschlimb:

### DIFF
--- a/docs/rfcs/20220804-ptensor/README.md
+++ b/docs/rfcs/20220804-ptensor/README.md
@@ -1,0 +1,66 @@
+# RFC: Parallel Tensor Dialect (PTensor)
+## Summary
+We propose a solution which can provide a high-level dialect for tensors and (numpy-like) tensor-operations allowing passes to create parallelism for GPU, OpenMP but also distributed memory while following the compute-follows-data concept.
+
+## Motivation
+MLIR provides a few tensor dialects (like tensor and TOSA) and separate dialects/conversions for targeting devices and parallelism. More complicated flows, for example when device and host tensors exist at the same time, complicated analysis is needed to correctly implement compute-follows data. Similarly, converting to distributed parallelism, such as MPI-based, is difficult to get right with existing dialects if possible at all.
+Additionally, the current architecture in Plier does not have an explicit MLIR representation of arrays/tensors and so makes it very hard to re-use its tensor functionality without coming from python/numba. Currently conversions are even relying the Python runtime. A dialect for such numpy-like tensors and -operations can accept hints for targets (like devices, MPI etc) and enable its use in libraries independent of numba and Python.
+
+## Proposal
+We propose two new Dialects:
+1. __PTensor__: providing a tensor type (ptensor) and operations on such ptensors
+2. __Dist__: dealing with the aspects of distributed data management, such as distributed memory alloction, GC and communication
+3. FIXME: I hope we do not need an extra GPU dialect and can use what already exists
+
+Additionally we propose appropriate passes
+1. Converting __PTensor__ operations into dialects __Linalg__, __GPU__ and __Dist__
+2. Converting __Dist__ operations into appropriate runtime calls
+
+### ptensor Type
+The ptensor type extends the `mlir::tensor` with `device` and `dist` attributes.
+The `device` indicates where the tensor lives, e.g. on which device.
+The target device is represented as a plain string which will be forwarded to the gpu/distributed runtimes. The exact syntax of device-identifiers is defined by the GPU runtime. The default value is `device="default"`.
+The `dist` attribute indicates whether the tensor is partitioned-distributed or not (boolean `UnitAttr`) and its default is `dist=false`.
+
+### ptensor Operations
+The initial set of operations matches the requirements of the core of [array-API](https://data-apis.org/array-api/latest/API_specification/index.html). Notice, since we focus in compute-follows-data, only the creation functions/operations will require the `device` and `dist` attributes. Operations consuming ptensors react on the the `device` and `dist` attribute of their input.
+
+### dist Operations
+- `alloc_dtensor(shape) : (ValueRange) -> (int64, RankedTensor)`
+- `dealloc_dtensor(dtensor) : (int64) -> void`
+- `create_view(dtensor, slice) : (int64, slice) -> (int64, RankedTensor)`
+- `local_shape(dtensor) : (int64) -> shape.shape`
+- `finalize_reduce(dtensor) : (int64) -> void`
+- `copy_view(from, to) : (int64, int64) -> void`
+
+### Passes
+All passes which consume `ptensor`s and -operations comply to compute-follows-data: Generation of GPU-, distribution-, parallelization- et al. operations depend on the `dist` and `device` attributes of input tensors. For example, if a tensor has the attribute "device='GPU'" then memory allocation operations must target the GPU.
+
+#### --lower-ptensor
+This pass completely lowers ptensor operations to
+- __Tensor__: Tensor types will convert to plain tensors. `dist` and `device` attributes get stripped off and used to appropriately create necessary operations in __GPU__ and __Dist__ dialects.
+- __Linalg__: The actual functionality will be represnted by one or more operations of the Linalg dialect.
+- __GPU__: highest level GPU dialect to indicate what needs to go to GPUs
+- __Dist__: new dialect which deals with primitives to manage operations on distributed tensors. Similar to the __GPU__ dialect the primitives might lower to interaction with a runtime (library).
+- utility dialects like __shape__, __affine__, __func__ and __arith__
+
+Combining conversions to multiple dialects keeps analysis simple and allows effective code generation. Some operations require interleaving operations from various dialects. Separate passes would become unnecessarily complicated because they need to understand the context (which we know while we convert ptensor). For instance, distributed `arange` requires various interactions with __Dist__ interleaving process-local operation in __Linalg__; a simple wrap of the local operation is not feasible.
+
+Complicated computation kernels might be simply replaced with a an appropriate library call to MKL or alike.
+
+Possible parameters to this pass could be flags to ignore `dist` and/or `device` attributes.
+
+#### --dist-to-intel-runtime
+FIXME: name and flow should be similar to what happens on the GPU side
+This pass converts all operations from __Dist__ into appropriate calls to the "Intel distributed runtime for MLIR (C)". This library provides the necessary hooks to deal with aspects like data/shape partitioning, tensor registration, GC, communication and more.
+
+## Alternatives
+We could also consider the `TOSA` dialect, either by extending it directly or lowering to it instead of to Linalg. The benefit of the latter is unclear and the first seems reasonable only if we can get the MLIR community's/TOSA maintainers' buy in.
+
+## Remarks
+- most of the __Dist__-unrelated work is planned independent of this in some form or another (elminating dependence on Python runtime, streamlining compile pipeline) or can be copied from existing passes in Plier. PTensor and its passes will provide a structured and re-usable context for them.
+- The detailed API of the distributed runtime is to be defined.
+
+## Questions
+- target: mapping queue <-> device
+- how are differences in HW handled in GPU pipeline/SPIR-V?


docs/rfcs/20220804-ptensor/README.md
```diff
@@ -0,0 +1,215 @@
...
+### __CFD__ Operations
+The CFD dialect provides region operation which defines the target on which operations within the region should be executed on. The region is configureted with an attribute which defines the HW type, such as 'dist' for distributed, or 'CUDA', 'L0', and alike for GPU execution. The exact syntax and configuration possiblities depend on the targeted runtime.
+Regions can be nested.
+Additionally, regions accept an optional runtime parameter
+* a team identifier for distributed regions refering to a team of processes
+* a device identifier for GPU regions refering to a specific HW device
+
+Example nesting a GPU region into a distributed region ('IDR'= Intel Distributed Runtime)
+```
+cfd.region{'IDR'}(0) {
```
Hardcode84, line 143: I don't think 'CFD' is a good name for this, as is is not tied specifically to `compute follows data` concept, and op itself should better be named something like `device.region`

fschlimb: What do you suggest instead of 'cfd'?

fschlimb:
@Jianhui-Li @petterinteon

docs/rfcs/20220804-ptensor/README.md
Hardcode84: At least in main I'm going to put it into `util` dialect as 'util' is not going away any time soon.

fschlimb: 'util' does not seem to be any better, so I assume you do not suggest this for refactor. In clean new world, what name would you prefer?

docs/rfcs/20220804-ptensor/README.md
```diff
@@ -0,0 +1,225 @@
...
+
+#### Operation details
+* Tensor creation
+  * `arange(start, stop, step, dtype, device) : (int64, int64, int64, type, str) -> ptensor.ptensor`
+  * `asarray(??) : (??) -> ptensor.ptensor`
+  * `create(shape, value, dtype, device) : (shape.shape, anytype, type, str) -> ptensor.ptensor`
+    * covers `empty, ones, zeros, full`
+  * `create_like(rsh, value, dtype, device) : (shape.shape, anytype, type, str) -> ptensor.ptensor`
+    * covers `empty_like, ones_like, zeros_like, full_like`
+  * `eye(n_rows, n_cols, k, dtype, device) : (int63, int64, int64, type, str) -> ptensor.ptensor`
```
diptorupd, line 66: typo `s\int63\int64\g`

docs/rfcs/20220804-ptensor/README.md
diptorupd: Having the region op is a clean way to annotate the device on which a ptensor op will execute. 

But, I am not in favour of separate a `CFD` dialect. I feel the op should be inside the `Dist` dialect and `Intel GPU/Device/XPU` dialect. Something like:
```
dist.distributed_region
ixpu.device_region
```
The names are indicative and we can look for alternatives.

However, if these ops cannot be made part of `Dist` and  `Intel GPU/Device/XPU`, then the op should be part of `ptensor`.

docs/rfcs/20220804-ptensor/README.md
fschlimb: fixed

docs/rfcs/20220804-ptensor/README.md
fschlimb: thanks @diptorupd @Hardcode84 I pushed a new revision without a cfd dialect and no dist_region.

docs/rfcs/20220804-ptensor/README.md
```diff
@@ -0,0 +1,268 @@
...
+2. __Dist__ dialect: dealing with the aspects of distributed data management, such as distributed memory alloction, GC and communication
+3. __intel_sycl.device_region__ operation in __intel_sycl__ dialect (RFC #291)
+
+
+Additionally we propose appropriate passes
+1. Converting __PTensor__ operations into dialects __Linalg__, __GPU__, __intel_sycl__ and __Dist__
+2. Converting __Dist__ operations into appropriate runtime calls
+3. Converting __intel-sycl.device_region__ to appropriate runtime calls.
+
+### ptensor Type
```
Hardcode84, line 26: We still need to decide (and document) if want value semantics like upstream `tensor` type, where each tensor is immutable, or reference semantics like upstream `memref` which allow inplace modifications. Most likely we need reference, to properly support numpy array assignment.

Hardcode84, line 26: Another topic if we want `ptensor` dialect to have explicit dealloc operations. Upstream `tensor` doesn't have them, considering them too `low-level` for its level of abstraction, but `memref` does have them and there are passes to insert deallocs automatically into memref code.

fschlimb: Good point.

I can't find anything in the docs saying that the builtin 'tensor' is immutable. The tensor dialect actually has 'insert' and 'insert_slice' opterations.

In any case, my latest commit clarifies that ptensor is intended to be mutable and to lower to memref. This is aligned with Linalg.

fschlimb: Good point. Yes, I think we need a way to dealloc. For dist operation there will be some kind of GC on remote processes which will have to trigger deallocation explicitly.

diptorupd:
@sklam tagging Siu from Numba to get his feedback from Numba's perspective.

docs/rfcs/20220804-ptensor/README.md
Hardcode84: > dist.distributed_region
ixpu.device_region

I don't think having multiple versions of this op in different dialects is a good idea. We will need to copypaste a lot of canonicalization and support code between them.

Hardcode84: Also, I'm considering to use op for CPU side parallelism too (prange) but I didn't decided yet.

fschlimb: Please note: the latest revision no longer has a distributed region. The region is mostly useful to carry the device information to lower passes and so enable cfd. We don't need this for dist.

fschlimb:
@pspillai

fschlimb:
@silee2 I included your suggestion to have ops support not only ptensors but also tensors and memrefs. Results will still be ptensors.

Hardcode84:
I still think ptensor `device` should be generic attribute, not limited to string, and also `team` and `gid` attributes should be part of `device`, as in most usecases (CPU/GPU) they will be unused and will just confuse users.

fschlimb:
> I still think ptensor `device` should be generic attribute, not limited to string, and also `team` and `gid` attributes should be part of `device`, as in most usecases (CPU/GPU) they will be unused and will just confuse users.

I am good with having device be 'AnyType'. We talked about unifying these into into and I recall we could not come up with a satisfying way of doing that. Do you have a suggestion how this could look like (and work) in practice?

Hardcode84:
> Do you have a suggestion how this could look like (and work) in practice?

By introducing custom attribute type in dialect.

fschlimb:
> > Do you have a suggestion how this could look like (and work) in practice?
> 
> By introducing custom attribute type in dialect.

The intent is to lower to something standard in the lower-ptensor pass. With that in mind, can you explain how this would make a difference?

Hardcode84:
> The intent is to lower to something standard in the lower-ptensor pass. 

 I don't see how it is relevant.

My suggestion is to represent ptensor type as `ptensor<2x3xf32, device<type:gpu1, team:whatever, gid:whatever>>` instead of `ptensor<2x3xf32, device:gpu1, team:whatever, gid:whatever>`. Because most cases will be non-distributed and `team` on type will be just unnnessesary and confusing, and for them we want something simple like `ptensor<2x3xf32, device:gpu1>`.

 Also, in general it will be much more useful if you start to write how your dialect will look like in IR instead of plain text.

fschlimb:
> My suggestion is to represent ptensor type as `ptensor<2x3xf32, device<type:gpu1, team:whatever, gid:whatever>>` instead of `ptensor<2x3xf32, device:gpu1, team:whatever, gid:whatever>`. Because most cases will be non-distributed and `team` on type will be just unnnessesary and confusing, and for them we want something simple like `ptensor<2x3xf32, device:gpu1>` instead of `ptensor<2x3xf32, device:gpu1, team:none, gid:none>`.

I think you are suggesting to make 'device', 'team' and 'gid' optional. This makes a lot of sense to me. Can you explain the benefit of introducing yet another type combining these attributes over having them individually as optional attributes in the ptensor type?

Hardcode84:
Most of the `ptensor` uses cases will be outside distributed context and having those very specific and low level attributes will look like a hack and will be confusing to most of the `ptensor` users. I don't see any issue introducing custom attribute type which can be set as `ptensor` device type and this new type can live in its own `dist` dialect.

fschlimb:
> Most of the `ptensor` uses cases will be outside distributed context and having those very specific and low level attributes will look like a hack and will be confusing to most of the `ptensor` users. I don't see any issue introducing custom attribute type which can be set as `ptensor` device type and this new type can live in its own `dist` dialect.

I am not arguing against your reasoning (as mentioned above, I agree). With my question I am only trying to figure out if optional attributes in the PTensor type wouldn't address your concern in a simpler way (e.g. avoiding an indirection). Can you explain the benefit of a separate type?

Hardcode84:
Even if we can hide optional attributes in asm form, they will still be there in generated C++ API to confuse users and add unnecessary clutter.

docs/rfcs/20220804-ptensor/README.md
```diff
@@ -0,0 +1,260 @@
...
+
+The optional `team` attribute indicates a team (of processes) among which the tensor is partitioned-distributed. The type of the `team` attribute depends on the underlying runtime; for MPI and Intel's distributed runtime this would be a `int64`. It defaults to `NoneType` which disables support for distributed operation.
+
+Notice: By default device and distribution support is disabled and so renders conventional host operations.
+
+### __PTensor__ Operations
+The initial set of operations matches the requirements of the core of [array-API](https://data-apis.org/array-api/latest/API_specification/index.html). Notice, since we focus on compute-follows-data, only the creation functions/operations will require the `device` and `team` attributes. The operations in the PTensor dialect accept different tensor-like types as input arguments: ptensors, tensors and MemRefs. Apparently, operations can only implement compute-follows-data if the inputs have `device` and `team` attributes, e.g. if they are of ptensor type. Standard tensor and memref inputs are treated as ptensors with default `device` and `team`.
```
Hardcode84, line 35: Actually, I don't like the idea of ptensor ops accepting usual tensors/memref arguments, better to have separate `to/from_tensor` and `to/from_memref` ops, which can also be useful outside this context.

fschlimb: Good idea.

I understand fromTensor and from fromMemref. The other way round raises some questions. If the ptensor is a device tensor and/or distributed, how would suggest that toMemref/toTensor behave? Raise an error?

Maybe we do not need tomemref/tensor at all.

fschlimb:
> Even if we can hide optional attributes in asm form, they will still be there in generated C++ API to confuse users and add unnecessary clutter.

I don't see that. The optional attributes will be somewhere no matter what. The clutter will be somewhere. We can easily provide creation/builder functions to hide the 'clutter'. In a pass accepting ptensors the device/dist info needs to be understood no matter what. I don't think the existence of a few additional accessor method will confuse anyone. However, if we export the attributes to another type we add complication (in use and documentation).

docs/rfcs/20220804-ptensor/README.md
Hardcode84: I don't see any problem in this, we want to lower ptensor to usual tensor/memref eventually anyway, and at this point those ops will be just noop. We can left this question up to the lowering passes, so they can choose, if they want to lower them or just return error.

Hardcode84:
> I don't see that. The optional attributes will be somewhere no matter what. The clutter will be somewhere.

In this case It will only seen by dist dialect users and passes. Everyone else who doesn't care about distributed stuff won't see this.

fschlimb:
@silee2 @diptorupd I clarified the broadcasting semantics. Different from what we said last time, the broadcasting needs to be implicit. Explicit broadcasting 'before' ptensor will likely hinder potential optimizations downstream. There are only few cases where implicit broadcasting is needed anyway. We operate on ranked tensors only (except 'squeeze' which requires an shaped tensor, even).

docs/rfcs/20220804-ptensor/README.md
fschlimb: Yes, we do not need to provide an implementation. Still, we need to define the semantics of the operation. The input to these operation is a ptensor, and the general expectation in this RFC is compute-follows-data. ToMemref and toTensor  effectively remove the location information. So what does this mean? The result lives in the default location (and so potentially needs to get copied over)? At least to me this does not make sense. And I do not see any other reasonable behavior.

If we can't define the semantics we better not have tomemref or totensor.

fschlimb:
> > I don't see that. The optional attributes will be somewhere no matter what. The clutter will be somewhere.
> 
> In this case It will only seen by dist dialect users and passes. Everyone else who doesn't care about distributed stuff won't see this.

OK. Relative to the rest of the RFC this is probably an implementations detail. I re-formulate the respective sections in a way that both solutions are possible. We can revisit this in respective PRs later.

docs/rfcs/20220804-ptensor/README.md
Hardcode84: Ok, I don't need them now for my cases. We can decide and add them later if needed.

fschlimb: OK.

I added a `from_ranked` operation and changed the description accordingly.

silee2 (approved):
LGTM.
@fschlimb, as you wrote in this RFC, details of dist dialect will be subject of another RFC.
Other parts look great. Thanks for all the hard work you have put into setting a good example of how future RFCs should be written.

fschlimb:
I will now start issuing PRs for getting existing code on refactor conforming to this RFC.
Such PRs might include clarifications/modifications to this RFC.

PR #284, opened 2022-08-07
Add additional canonicalization to ChangeLayoutOp

Hardcode84:
`ChangeLayoutOp` is used to 'align' memrefs layouts after bufferization. Canonicalization must propagate layout through ops (e.g 'change_layout -> op' must be transformed to 'op -> change_layout').
* Add transformation canonicalization for `arith.select`
* Add custom assembly format to `LayoutChangeOp`
* Add tests for `arith.select` and some other ops
* Reenable affected test

### DIFF
--- a/mlir/lib/Dialect/plier_util/dialect.cpp
+++ b/mlir/lib/Dialect/plier_util/dialect.cpp
@@ -1041,6 +1041,56 @@ struct ChangeLayoutExpandShape
   }
 };
 
+struct ChangeLayoutSelect
+    : public mlir::OpRewritePattern<mlir::arith::SelectOp> {
+  using OpRewritePattern::OpRewritePattern;
+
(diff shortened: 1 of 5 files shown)


Hardcode84:
ping

mlir/lib/Dialect/plier_util/dialect.cpp
```diff
@@ -1041,6 +1041,57 @@ struct ChangeLayoutExpandShape
   }
 };
 
+struct ChangeLayoutSelect
```
fschlimb, line 1044: Please add comments, there is now way I can review without.

Hardcode84: done

Hardcode84:
Sorry, but this is too slow for a trivial fix, I'm merging this

PR #285, opened 2022-08-08
Additional includes to stdlib.h to enable compile using gcc11

petterinteon:


dpcomp_runtime/lib/memory.cpp
```diff
@@ -14,6 +14,7 @@
 
 #include <cstdint>
 #include <cstring>
+#include <stdlib.h>
```
fschlimb, line 17: Better `#include <cstdlib>`

petterinteon: Fixed.

dpcomp_runtime/lib/memory.cpp
```diff
@@ -14,6 +14,7 @@
 
 #include <cstdint>
 #include <cstring>
+#include <cstdlib>
 #if defined(__MACH__)
 #include <cstdlib>
```
Hardcode84, line 19: this can be removed now

fschlimb (approved):

PR #288, opened 2022-08-10
Remove custom pass ParallelLoopGPUMappingPass

nbpatel:

We dont require this pass. Upstream has a pass which does the mapping.

### DIFF
--- a/mlir/tools/imex-opt/passes.cpp
+++ b/mlir/tools/imex-opt/passes.cpp
@@ -29,13 +29,6 @@
 // Passes registration.
 
 static mlir::PassPipelineRegistration<>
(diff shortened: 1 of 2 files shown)


Hardcode84:
It broke python gpu tests

mlir/tools/imex-opt/passes.cpp
```diff
@@ -28,13 +28,6 @@
 
 // Passes registration.
 
-static mlir::PassPipelineRegistration<>
```
Hardcode84, line 31: please, do no remove it from imex-opt as I'm planning to write lit test for it, LGTM othertwise

Hardcode84 (approved):

nbpatel:
#318

PR #291, opened 2022-08-15
Initial Draft for the I-GPU Dialect RFC

nbpatel:

This PR provides the initial draft for the Intel Gpu dialect. The Intel GPU dialect is a minimal dialect which replaces the gpu runtime dialect which has a lot of duplicated ops in it. This initial PR is an RFC.

### DIFF
--- a/I-GPU-RFC.md
+++ b/I-GPU-RFC.md
@@ -1,0 +1,44 @@
+# RFC: Intel GPU Dialect (I-GPU)
+
+Nishant Patel 
+
+## Summary
+
+We propose the Intel Gpu Dialect as a custom Intel dialect to unlock the underlying runtimes (Level zero & Sycl) best suited for Intel Gpus.
+This Dialect will be at a lower level than the upstream GPU Dialect. 
+
+## Motivation
+
+Upstream MLIR has a Dialect called the [GPU Dialect](https://mlir.llvm.org/docs/Dialects/GPU/) which provides middle-level abstractions for launching GPU kernels following a programming model similar to that of CUDA or OpenCL. The upstream dialect exposes all the operation required to launch the kernel on a GPU device. But, the drawback with the upstream dialect is that it is designed majorly catering to CUDA. One of the problem with upstream dialect is that it does not expose Stream as an operation in the dialect. For the upstream dialect, stream is created internally while lowering to LLVM runtime calls and it refers specifically to the CUDA stream. Therefore, the need to create an I-GPU Dialect arises. The I-GPU dialect is an extension of the upstream GPU dialect, with a single op called GetStreamOp. The GetStreamOp op is required to prepare the Intel GPU stream for underlying runtime (SYCL/L0). Stream, here, can be thought of as a structure which is similar to sycl queue in Sycl runtime and command queue in L0 runtime. Exposing stream as an operation is necessary to have more control over the underlying runtime and also enables us to take the stream as an input provided by some external user. 
+
+## Proposal
+
+We propose to have the I-GPU dialect with just one op called the GetStreamOp.
+
+### Operation:
+
+#### intel_gpu.get_stream (intel_gpu::GetStreamOp)
+
+Intel GPU Stream Operation
+
+Syntax:
+
+```
+operation ::= "intel_gpu.get_stream"() : () -> !intel_gpu.OpaqueType
+```
+
+The intel_gpu.get_stream() operation will be added by a pass called IntelGpuStreamOp after the dialect is in Gpu Dialect. After that the IR is lowered all the way down to LLVM.
+This operation will be converted to runtime call in the LLVM generated code. The runtime routine will create a SYCL/L0 queue. Queue is used to schedule operations on the device (GPU). From the queue, one can extract the device and context. 
+
+
+## Alternative
+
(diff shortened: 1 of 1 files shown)


Hardcode84:
Sorry, but don't see much value in this RFC.
I introduced our custom GPU ops because existing GPU dialect was missing following features:
* Explicit stream/context creation
* Explicit module/kernel loading ops
And it will be much more productive to propose and discuss these changes upstream instead of discussing them here.

nbpatel:
@Hardcode84 The custom GPU ops you have are all duplicated from upstream GPU dialect and we don't see any need of duplicating those hence we propose this minimal dialect with just one op.

Hardcode84:
`get_stream` doesn't make any sense without other GPU ops taking stream as argument. With single `get_stream` op this RFC just plan useless.

nbpatel:
Other GPU ops take this stream as an argument. But they are the upstream GPU ops, there is no need to duplicate them in our own dialect.

Hardcode84:
> Other GPU ops take this stream as an argument.

They do not. If you mean `AsyncDependencies` param, it is there for completely different purpose.

nbpatel:
not the asyndependencies param, you can change the upstream lowering pass to take a custom stream rather than the stream the upstream lowering uses.

fschlimb:
Thanks @nbpatel, this looks good.

A few (mostly minor) remarks:
* I suggest you setup the pre-commit hook in your local repo as described in the README. With this, any commit will get auto-checked and in most cases even auto-corrected with respect what the pre-commit  check requires.
* Maybe you want to consider moving some of the details in "motivation" to the actual proposal. To me the separation of motivation and proposal could be a bit clearer.
* I don't fully understand the lowering flow. Could you add a few more details (including details about the above issue raised by @Hardcode84)?
* For my own understanding:  It seems the new operation is generated out of nothing, how would the compiler tell the runtime which device to create a queue/context for? Will it just use any/default device?

I-GPU-RFC.md
```diff
@@ -0,0 +1,44 @@
...
+operation ::= "intel_gpu.get_stream"() : () -> !intel_gpu.OpaqueType
+```
+
+The intel_gpu.get_stream() operation will be added by a pass called IntelGpuStreamOp after the dialect is in Gpu Dialect. After that the IR is lowered all the way down to LLVM.
+This operation will be converted to runtime call in the LLVM generated code. The runtime routine will create a SYCL/L0 queue. Queue is used to schedule operations on the device (GPU). From the queue, one can extract the device and context. 
+
+
+## Alternative
+
+The alternative approch here is to add this operation in the upstream dialect and extend the upstream GPU dialect. But, we feel that having our own dialect will give us more command and control over how we want to design it. Nvidia and AMD both have their own custom dialect at a similar level.
```
diptorupd, line 36: > Nvidia and AMD both have their own custom dialect at a similar level.
Are you referring to `amdgpu` and `nvgpu`? Can you expand on the statement and provide examples on how the Intel GPU dialect compares to the AMD and Nvidia specific GPU dialects?

I-GPU-RFC.md
```diff
@@ -0,0 +1,44 @@
+# RFC: Intel GPU Dialect (I-GPU)
```
diptorupd, line 1: I do not like the name Intel GPU Dialect. SYCL and L0 do not just GPU devices. The name should be something more generic.

diptorupd:
@nbpatel can you add an example showing the intended usage of the dialect? Especially how it is meant to be used in conjunction with the upstream GPU dialect.

@mehdi-goli Your feed back will be useful.

diptorupd:
> Please review these guidelines to help with the review process:
> 
>     * [ ]  Have you provided a meaningful PR description?
> 
>     * [ ]  Have you added a test, a reproducer, or a reference to an issue with a reproducer?
> 
>     * [ ]  Have you tested your changes locally for CPU and GPU devices?
> 
>     * [ ]  Have you made sure that new changes do not introduce compiler warnings?
> 
>     * [ ]  If this PR is a work in progress, are you filing the PR as a draft?
> 
>     * [ ]  Have you organized your commits logically and ensured each can be built by itself?

@nbpatel *Please* add a meaningful description to your PR. The checkboxes are meant to prod you in that direction :) not be a substitute for a description.

nbpatel:
Thanks @diptorupd and @fschlimb for the comments. I will add more details and example as Dipto mentioned. Also, @fschlimb for device/context, as of now, it uses a default device/context as you mentioned. I have it in my questions section if we will need separate ops for those as well.

I-GPU-RFC.md
nbpatel: since the amd and nvidia dialects are called  AMDGPU and NVGPU dialects, I thought Intel-GPU dialect made sense. We can discuss though. I will add more details about those dialects in the RFC

fschlimb:
> Also, @fschlimb for device/context, as of now, it uses a default device/context as you mentioned. I have it in my questions section if we will need separate ops for those as well.

What about including the device region op as discussed in #283 ?

mehdi-goli:
> @nbpatel can you add an example showing the intended usage of the dialect? Especially how it is meant to be used in conjunction with the upstream GPU dialect.
> 
> @mehdi-goli Your feed back will be useful.

Hi @diptorupd , I don't think the SYCL queue and level zero command group can completely relate in this case. I think the target here is the Level zero command group. In this case, having the stream helps to define the dependency between the memory, especially for a graph-based model when multiple streams have been used.

nbpatel:
will create a new PR with changes and the file in the correct rfc folder

PR #292, opened 2022-08-18
Introduce device/environment region op

Hardcode84:
* Introduce `EnvironmentRegionOp` and corresponding `EnvironmentRegionYieldOp` terminator.
* Renamed from previous `device_region` proposal to `env_region` as it is not strictly tied to the device (e.g. GPU) execution.
* This PR only introduces op definition and some basic printing/parsing tests.
* For the specific op semantics see op description in td file.


fschlimb (commented):
Is there a way to have the printing and parsing code get auto-generated?

Hardcode84:
> Is there a way to have the printing and parsing code get auto-generated?

Not for the ops which manipulate regions as far as I know.

Hardcode84:
Can we get review on actual semantics? It is blocking further work, and we don't have any comment on semantics I would prefer it to be merged and any potential minor issues like actual text formatting fixed postfactum.

fschlimb:
> Can we get review on actual semantics? It is blocking further work, and we don't have any comment on semantics I would prefer it to be merged and any potential minor issues like actual text formatting fixed postfactum.

We have a RFC #283 which is intended to define the semantics. You are welcome to call a separate RFC if you prefer a separate document for this.

Hardcode84:
This is much better than RFC, because it shows actual code, which can be compiled and run. But you can consider op description in td file as 'RFC'.

Hardcode84:
If there aren't any further questions I'm going to merge this.

PR #296, opened 2022-08-26
Remove IMEX_ENABLE_NUMBA_HOTFIX

Hardcode84:
* gpu_runtime was the last place where `IMEX_ENABLE_NUMBA_HOTFIX` was used
* GPU tests locally pass without it

### DIFF
--- a/dpcomp_gpu_runtime/lib/gpu_runtime.cpp
+++ b/dpcomp_gpu_runtime/lib/gpu_runtime.cpp
@@ -346,9 +346,7 @@ struct Stream {
       throw std::runtime_error("Failed to allocate MemInfo");
     }
 
-#if defined(IMEX_ENABLE_NUMBA_HOTFIX)
     retain();
-#endif
     return {info, mem, event};
   }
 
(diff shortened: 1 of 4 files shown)


Hardcode84:
@nbpatel What was the exact issue here?

GPU tests pass locally (except one double test). Can you test it on your side?

dpcomp_gpu_runtime/lib/gpu_runtime.cpp
```diff
@@ -346,9 +346,7 @@ struct Stream {
       throw std::runtime_error("Failed to allocate MemInfo");
     }
 
-#if defined(IMEX_ENABLE_NUMBA_HOTFIX)
     retain();
```
nbpatel, line 349: this was throwing an exception IIRC, I think this is used for reference counting for your python stuff and we dont require it

Hardcode84: It should only cause memory leak and not crash, GPU tests pass locally for me, can you check it on your side?

nbpatel: yes, it gives a memory leak

Hardcode84:
Changed gpu buffer handling so `Stream` object should no longer leak as long as allocated buffers doesn't leak (have corresponding `gpu.dealloc`)

nbpatel (approved):
LGTM. Please make sure the gpu tests pass

Hardcode84:
They pass locally for me, can you check on you side too?

PR #306, opened 2022-08-31
Add support for building IMEX as part of LLVM build with

silee2:
-DLLVM_EXTERNAL_PROJECTS

For example if you have both llvm and imex source code directory side by side
$HOME/llvm-project
$HOME/mlir-extensions

You can configure both llvm and imex with the following
cd $HOME/llvm-project
cmake -B build -G Ninja -S llvm \
   -DCMAKE_CXX_STANDARD=17 \
   -DLLVM_ENABLE_PROJECTS=mlir \
   -DLLVM_BUILD_EXAMPLES=ON \
   -DLLVM_TARGETS_TO_BUILD="X86" \
   -DCMAKE_BUILD_TYPE=Release \
   -DLLVM_ENABLE_ASSERTIONS=ON \
    -DLLVM_EXTERNAL_PROJECTS="Imex" \
    -DLLVM_EXTERNAL_IMEX_SOURCE_DIR="../mlir-extensions"

And after config, can build and run imex tests with the following
cmake --build build
cmake --build build --target check-imex


fschlimb:
Nice!

I guess this is great for people who want to try IMEX. Personally I prefer providing build instructions like the above over providing custom build scripts (of course only as long as they are simple enough).

Building IMEX already takes relatively long time even though there is almost nothing in refactor. How long does it take cmake-build to fast-forward to IMEX build if LLVM/MLIR has not changed?

Where will the IMEX install go to with this mechanism? An isolated path would make it easier to integrate this into our CI. As you rename the doc-file for the conversion/general passes I conclude that at least docs will be fully integrated into the MLIR docs structure. For other files (like include files) such a integration will make it harder to cache a LLVM build in CI.

I think we are fine at compile time (for projects using IMEX) as long as IMEX include files continue to live under 'include/imex'.

A few minor comments/thoughts:
* We should use/recommend -DLLVM_INSTALL_UTILS=ON when building llvm/mlir
* The README should include the instructions
* I assume a 'distribution' of imex would still be stand-alone, e.g. come with its own cmake configs etc.

@diptorupd might have additional thoughts

drprajap (approved):
LGTM,  I had same question as Frank on Install utils, is there specific reason you have not included in build instructions?

silee2:
Updated README.md with reason for not including -DLLVM_INSTALL_UTILS=ON

silee2:
This PR needs to be merge after #330 since LLVM will default to C++17 after that PR.

Dewei-Wang-sh (approved):
lgtm

PR #310, opened 2022-09-05
Port InsertGpuAlloc, SetSPIRVCapabilities & SetAbiAttrPass from main

nbpatel:

This PR ports the InsertGpuAllocs, SetSPIRVCapabilities & SetAbiAttrPass from main branch. The passes have test cases and a doc attached to them.


Hardcode84 (changes requested):
Sorrym but we need all these tests in main first

fschlimb (changes requested):
Wow, we are getting some real functionality into refactor! God to see comprehensive (user-)docs and tests!

We should follow the upstream conventions and have the docu in the .td files (yes, hey have some separate files in docs, but that's the exception). If we follow this process we get the standard layout, formatting and a TOC maintained for free.

One reason for the refactor was to make sure that all code is in a state which allows non-authors to easily understand it. The code here has almost no comments, that's not good. I know this is how it is on main. However here, at least a summary of the functional flow in each file as well as of each (non-trivial) function should be added. I don't think a meaningful review can be done without.

fschlimb:
@nbpatel if you install the "precommit" to your local repository (from where you push) - as mentioned in the RADME - you will get most of the formatting issues reported by pre-commit-check automatically corrected during your commit there. If you have issues using it, we should fix it (you might raise an issue for that).

tools/imex-opt/imex-opt.cpp
```diff
@@ -38,3 +39,22 @@ int main(int argc, char **argv) {
   return ::mlir::asMainReturnCode(
       ::mlir::MlirOptMain(argc, argv, "Imex optimizer driver\n", registry));
 }
+
```
silee2, line 42: Would be better if registration could be done through "::imex::registerAllPasses()" automatically by using files like Passes.td Passes.h but could be done in a follow up PR.

silee2 (approved):
LGTM

lib/Transforms/InsertGpuAllocs.cpp
```diff
@@ -0,0 +1,363 @@
...
+                  mem = parentView.getViewSource();
+
+                for (auto alias : aliases.resolve(mem)) {
+                  auto op = alias.getDefiningOp();
+                  if (op) {
+                    if (mlir::isa<mlir::memref::GetGlobalOp>(op)) {
+                      gpuGetMemrefGlobalParams.insert({op, {}});
+                      continue;
+                    }
+                    if (op->getDialect() == scfDialect ||
```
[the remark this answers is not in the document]
Jianhui-Li: Also why only scfDialect? There are so many other dialects. like linalg/Affine loop could produce memref also?

lib/Transforms/InsertGpuAllocs.cpp
```diff
@@ -0,0 +1,363 @@
...
+                    if (mlir::isa<mlir::memref::AllocOp>(op)) {
+                      gpuBufferAllocs.insert({op, {}});
+                    } else if (mlir::isa<mlir::func::CallOp>(op)) {
+                      // Ignore
+                    } else {
+                      op->emitError("Unhandled memref producer");
+                      return mlir::WalkResult::interrupt();
+                    }
+
+                  } else {
```
Jianhui-Li, line 136: what is the condition for this? a "non" op?

lib/Transforms/InsertGpuAllocs.cpp
```diff
@@ -0,0 +1,363 @@
...
+                    auto index = static_cast<unsigned>(it - blockArgs.begin());
+                    gpuBufferParams.insert({index, {}});
+                  }
+                }
+              }
+
+              return mlir::WalkResult::advance();
+            })
+            .wasInterrupted()) {
+      signalPassFailure();
```
Jianhui-Li, line 150: If the pass fail, will the L0runner fail?  Does GPU/L0 always guarantee that memrefs used inside kernel are always accessible, no matter where the data is located?

lib/Transforms/InsertGpuAllocs.cpp
```diff
@@ -0,0 +1,363 @@
...
+      *this, "use-gpu-dealloc",
+      llvm::cl::desc("use gpu.dealloc for gpu allocated memrefs, "
+                     "memref.dealloc will be used otherwise"),
+      llvm::cl::init(true)};
+};
+
+} // namespace imex
+
+std::unique_ptr<mlir::Pass>
+imex::createInsertGPUAllocsPass(bool useGpuDealloc) {
```
Jianhui-Li, line 361: when user set this bool as true and when as false?

lib/Transforms/InsertGpuAllocs.cpp
```diff
@@ -0,0 +1,363 @@
...
+        builder.create<mlir::gpu::DeallocOp>(loc, llvm::None, allocResult);
+      else
+        builder.create<mlir::memref::DeallocOp>(loc, allocResult);
+    }
+  }
+
+  Option<bool> useGpuDealloc{
+      *this, "use-gpu-dealloc",
+      llvm::cl::desc("use gpu.dealloc for gpu allocated memrefs, "
+                     "memref.dealloc will be used otherwise"),
```
Jianhui-Li, line 354: what is the reason to use memref.dealloc for GPU allocated memrefs?

lib/Transforms/InsertGpuAllocs.cpp
```diff
@@ -0,0 +1,363 @@
...
+      auto alloc = mlir::cast<mlir::memref::AllocOp>(it.first);
+      it.second = getAccessType(alloc);
+    }
+
+    auto &block = funcBody.front();
+    for (auto &it : gpuBufferParams) {
+      auto param = block.getArgument(it.first);
+      it.second = getAccessType(param);
+
+      it.second.hostRead = true;
```
Jianhui-Li, line 207: Does this mean that the function always assumes the input/output is in the host side?   This is not TRUE for XLA usage, where the caller controls the runtime resource, so the input/output passed would be in device side for most of time. 

Are there a use case where there are two consecutive function calls, the first one produce a buffer for the second one to use, can the intermediate buffer stay in the device memory?  How the current code would be restructured to support that need?  

I don't mean that this PR needs to fix these limitations.  But would be useful if we list these considerations as future development opportunities.

lib/Transforms/InsertGpuAllocs.cpp
```diff
@@ -0,0 +1,363 @@
...
+      auto loc = alloc.getLoc();
+      builder.setInsertionPoint(alloc);
+      auto gpuAlloc = builder.create<mlir::gpu::AllocOp>(
+          loc, alloc.getType(), /*asyncToken*/ nullptr,
+          /*asyncDependencies*/ llvm::None, alloc.dynamicSizes(),
+          alloc.symbolOperands());
+      alloc->replaceAllUsesWith(gpuAlloc);
+      alloc.erase();
+      if (access.hostRead || access.hostWrite)
+        gpuAlloc->setAttr(imex::getAllocSharedAttrName(),
```
Jianhui-Li, line 293: why in this case, there is no  deallocop, like in the parameter and global case? "builder.create<mlir::gpu::DeallocOp>(loc, llvm::None, allocResult);"?

lib/Transforms/InsertGpuAllocs.cpp
```diff
@@ -0,0 +1,363 @@
...
+                  auto op = alias.getDefiningOp();
+                  if (op) {
+                    if (mlir::isa<mlir::memref::GetGlobalOp>(op)) {
+                      gpuGetMemrefGlobalParams.insert({op, {}});
+                      continue;
+                    }
+                    if (op->getDialect() == scfDialect ||
+                        mlir::isa<mlir::ViewLikeOpInterface>(op))
+                      continue;
+                    if (mlir::isa<mlir::memref::AllocOp>(op)) {
```
Jianhui-Li, line 127: What about handling variants of Allocop? 
[memref.alloc (::mlir::memref::AllocOp)](https://mlir.llvm.org/docs/Dialects/MemRef/#memrefalloc-mlirmemrefallocop)
[memref.alloca (::mlir::memref::AllocaOp)](https://mlir.llvm.org/docs/Dialects/MemRef/#memrefalloca-mlirmemrefallocaop)
[memref.alloca_scope (::mlir::memref::AllocaScopeOp)](https://mlir.llvm.org/docs/Dialects/MemRef/#memrefalloca_scope-mlirmemrefallocascopeop)
[memref.alloca_scope.return (::mlir::memref::AllocaScopeReturnOp)](https://mlir.llvm.org/docs/Dialects/MemRef/#memrefalloca_scopereturn-mlirmemrefallocascopereturnop)

lib/Transforms/InsertGpuAllocs.cpp
Jianhui-Li: If a func first alloc a temp memory and then release it?  We only replace the memref.alloc, not memref.dealloc? 
Don't we need negative test cases also?

lib/Transforms/InsertGpuAllocs.cpp
```diff
@@ -0,0 +1,363 @@
...
+    // This is the case where a memref.alloc op is directly converted to
+    // gpu.alloc
+    for (auto it : gpuBufferAllocs) {
+      auto alloc = mlir::cast<mlir::memref::AllocOp>(it.first);
+      auto access = it.second;
+      auto loc = alloc.getLoc();
+      builder.setInsertionPoint(alloc);
+      auto gpuAlloc = builder.create<mlir::gpu::AllocOp>(
+          loc, alloc.getType(), /*asyncToken*/ nullptr,
+          /*asyncDependencies*/ llvm::None, alloc.dynamicSizes(),
```
Jianhui-Li, line 288: Why the getType and size are different than the global and parameters?  

      for (auto i : llvm::seq(0u, rank)) {
        if (memrefType.isDynamicDim(i)) {
          auto op = builder.create<mlir::memref::DimOp>(loc, param, i);
          dims.push_back(op);
          filter.insert(op);
        }
      }
      auto allocType = mlir::MemRefType::get(
          memrefType.getShape(), memrefType.getElementType(),
          mlir::MemRefLayoutAttrInterface{}, memrefType.getMemorySpace());
      auto gpuAlloc = builder.create<mlir::gpu::AllocOp>(
          loc, allocType, /*asyncToken*/ nullptr,
          /*asyncDependencies*/ llvm::None, dims,
          /*symbolOperands*/ llvm::None);

lib/Transforms/InsertGpuAllocs.cpp
```diff
@@ -0,0 +1,363 @@
...
+      alloc.erase();
+      if (access.hostRead || access.hostWrite)
+        gpuAlloc->setAttr(imex::getAllocSharedAttrName(),
+                          builder.getUnitAttr());
+    }
+
+    // This is the case where the inputs are passed as arguments to the
+    // function. This code will add the IR for memeory allocation on the device
+    // with gpu.alloc and insert a memref.copy from host to device
+    for (auto it : gpuBufferParams) {
```
Jianhui-Li, line 300: Is it possible for global and parameter to share code as much as possible if most of handling are same?

Jianhui-Li (changes requested):

lib/Transforms/InsertGpuAllocs.cpp
```diff
@@ -0,0 +1,363 @@
...
+
+    auto getMemref = [](mlir::Operation *op)
+        -> llvm::Optional<mlir::SmallVector<mlir::Value, 4>> {
+      if (auto load = mlir::dyn_cast<mlir::memref::LoadOp>(op)) {
+        return {{load.memref()}};
+      } else if (auto store = mlir::dyn_cast<mlir::memref::StoreOp>(op)) {
+        return {{store.memref()}};
+      } else if (auto call = mlir::dyn_cast<mlir::func::CallOp>(op)) {
+        mlir::SmallVector<mlir::Value, 4> ret;
+        for (auto arg : call.operands()) {
```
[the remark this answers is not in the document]
nbpatel: because there might be a func.call whose operands would be memrefs. basically we are just looking for all the memref producers

lib/Transforms/InsertGpuAllocs.cpp
```diff
@@ -0,0 +1,363 @@
...
+            return true;
+        }
+      }
+      return false;
+    };
+
+    // Traverse through all the memory access ops under GPU launch Op
+    // and add device memory allocation appropriately.
+    if (func.walk([&](mlir::Operation *op) {
+              if (!op->getParentOfType<mlir::gpu::LaunchOp>())
```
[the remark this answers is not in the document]
nbpatel: how would it be any different after the kernels are outlined? currently this pass runs before the kernels get outlined. The drawback I see is making sure the gpu level tests have to have the gpu.launch blocks ... format...

lib/Transforms/InsertGpuAllocs.cpp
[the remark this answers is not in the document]
nbpatel: I guess because memrefs produced by scf dialect ops are not actual memrefs and it is just control flow

lib/Transforms/InsertGpuAllocs.cpp
nbpatel: this is the gpu params case,

lib/Transforms/InsertGpuAllocs.cpp
nbpatel: yes, currently the runner will fail if the pass fails

lib/Transforms/InsertGpuAllocs.cpp
nbpatel: the bool is true for our l0 runner case where we use the conventional gpu alloc and gpu dealloc, and false for the case Ivan had where they have a different custom memref descriptor for their memory management and. they use memref dealloc to deallocate the memory

lib/Transforms/InsertGpuAllocs.cpp
nbpatel: mentioned above

lib/Transforms/InsertGpuAllocs.cpp
nbpatel: Good catch. I will add the code for adding the gpu.dealloc in this case

lib/Transforms/InsertGpuAllocs.cpp
nbpatel: these are not handled as of now in this PR, only memref::allocOp is handled

lib/Transforms/InsertGpuAllocs.cpp
nbpatel: because here you directly have a memref alloc to convert to gpu alloc and global and params case you dont have it, you add the allocs and subsequent mem copy's

lib/Transforms/InsertGpuAllocs.cpp
nbpatel: yes, you are right, the function assumes that the input/output is in host side. we dont have a test case of the above mentioned case. maybe in the future I will do a subsequent PR to address this?

docs/Transforms/InsertGpuAllocs.md
```diff
@@ -0,0 +1,80 @@
+# InsertGpuAllocs Pass
+
+
+The InsertGpuAllocs pass, as the name suggests, inserts the gpu allocs in the IR. Memref alloc is an operation in the memref dialect that can be used to allocate the memory on the host side and or on the device side. The MLIR IR is a mix of host and device code.
+To distinguish between host side memory allocation and device side memory allocation, we convert all the memref.allocs that refer to device (gpu) side memory allocations and references, into gpu.alloc, which is an operation of the upstream GPU dialect. This distinction helps in lowering to llvm and calling the appropriate memory allocation operation at runtime.
+The pass traverses all the memref (load/store) operations inside the gpu launch op in the IR and checks for its aliases and its defining op. If the defining op is a memref.alloc op it replaces that op in the IR with gpu.alloc op, because all the operations under the gpu.launch op are device side computations and will execute on the device.
```
fschlimb, line 6: I suggest moving this last line into the file with the implementation. This is good example of the kind of high-level description of the flow of the implementation which I suggest to have in all files. A "user" of this pass does not care about such details, but the reader of the code greatly benefits from it.

docs/Transforms/InsertGpuAllocs.md
```diff
@@ -0,0 +1,80 @@
...
+    %7 = arith.addf %5, %6 : f32
+    memref.store %7, %memref_3[%4] : memref<8xf32>
+    gpu.terminator
+  }
+  %3 = memref.cast %memref_3 : memref<8xf32> to memref<*xf32>
+  call @printMemrefF32(%3) : (memref<*xf32>) -> ()
+  return
+}
+```
+
```
fschlimb, line 73: Examples are great. Can you provide a smaller example?

lib/Transforms/InsertGpuAllocs.cpp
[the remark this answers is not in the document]
fschlimb: This is something you could add as a comment :)

lib/Transforms/InsertGpuAllocs.cpp
```diff
@@ -0,0 +1,363 @@
...
+///
+//===----------------------------------------------------------------------===//
+
+#include <imex/Transforms/Transforms.h>
+
+namespace imex {
+
+mlir::StringRef getAllocSharedAttrName() { return "gpu.alloc_shared"; }
+
+struct InsertGPUAllocs
```
fschlimb, line 22: some summary in doxygen style would auto-generate docs.

lib/Transforms/InsertGpuAllocs.cpp
```diff
@@ -0,0 +1,363 @@
...
+      if (auto call = mlir::dyn_cast<mlir::func::CallOp>(op)) {
+        for (auto arg : call.operands()) {
+          if (arg.getType().isa<mlir::MemRefType>())
+            return true;
+        }
+      }
+      return false;
+    };
+
+    // Traverse through all the memory access ops under GPU launch Op
```
fschlimb, line 99: Since you asked: until here, there is no comment or anything saying what happens in the code.

lib/Transforms/InsertGpuAllocs.cpp
```diff
@@ -0,0 +1,363 @@
...
+      auto param = block.getArgument(it.first);
+      it.second = getAccessType(param);
+
+      it.second.hostRead = true;
+      it.second.hostWrite = true;
+    }
+
+    // GetMemrefGlobal Op Case:
+    // This is the case where the inputs are globals contants and accessed using
+    // memref.get_global op. This code will add the IR for memeory allocation on
```
fschlimb, line 213: typo: memory

lib/Transforms/SetSPIRVCapabalities.cpp
```diff
@@ -0,0 +1,61 @@
+//===- SetSPIRVCapabalities.cpp - SetSPIRVCapabalities Pass  -------*- C++
```
fschlimb, line 1: filename has a typo

test/Transforms/set-spirv-capability.mlir
```diff
@@ -0,0 +1,41 @@
+// RUN: imex-opt --set-spirv-capablilities %s | FileCheck %s
+
+module attributes {gpu.container_module} {
+
+// CHECK: module attributes {gpu.container_module, spv.target_env = #spv.target_env<#spv.vce<v1.0, [Addresses, Float16Buffer, Int64, Int16, Int8, Kernel, Linkage, Vector16, GenericPointer, Groups, Float16, Float64, AtomicFloat32AddEXT, ExpectAssumeKHR], [SPV_EXT_shader_atomic_float_add, SPV_KHR_expect_assume]>, #spv.resource_limits<>>} {
```
fschlimb, line 5: Any particular reason why this test is so big? Most of the code here is not checked, so why is it needed?
It might be worthwhile generally reducing the suggested tests. Minimal tests are easier to maintain and understand.

tools/imex-opt/CMakeLists.txt
```diff
@@ -8,4 +8,4 @@ set(LIBS
 add_llvm_executable(imex-opt imex-opt.cpp)
 
 llvm_update_compile_flags(imex-opt)
-target_link_libraries(imex-opt PRIVATE ${LIBS})
+target_link_libraries(imex-opt PRIVATE ${LIBS} IMEXTransforms)
```
fschlimb, line 11: Why is IMEXTransforms not in ${LIBS}. It should be.

lib/Transforms/InsertGpuAllocs.cpp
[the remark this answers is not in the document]
Jianhui-Li: >>because there might be a func.call whose operands would be memrefs. basically, we are just looking for all the memref producers

I am a bit confused by this statement.  Store doesn't produce memref. There are so many other dialects/ops producing memrefs.  Why we only care load/store/call op? What exact analysis and transformation we do in this pass for these three ops?  

getMemRef() is a very generic name, and the implementation just takes three op categories.  I think it should be modified to be a narrower name exactly describing what it is.

lib/Transforms/InsertGpuAllocs.cpp
Jianhui-Li: Agree. Some reasonable comments are needed. Usually the questions from readers/reviewers are good hints where a comment might be needed.

lib/Transforms/InsertGpuAllocs.cpp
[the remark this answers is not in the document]
Jianhui-Li: >>how would it be any different after the kernels are outlined?
I think that may involve cross device func and host code analysis, since the memref used inside device kernel needs to be gpu.alloc in the host code.  I would expect the code needs to handle launch_func.  I don't know how important it is so far for our use cases, but I think we need to tell user upfront that we are not handling gpu.launch_func. 

>> currently this pass runs before the kernels get outlined.
Maybe you can put this in the pass introduction.  So user knows the limitation.

lib/Transforms/InsertGpuAllocs.cpp
Jianhui-Li: >>this is the gpu params case,
Here you may want to put an assert statement to assert your assumption. 

The logic is not straightforward - 

Here you assume that if alias.getDefiningOp() do return a null pointer,  then the assumption is that alias can give Parentblocks where you can find the arguments.  What if the loop body has nested loop inside? Do you write test case to verify?

lib/Transforms/InsertGpuAllocs.cpp
Jianhui-Li: Are these cases important to handle?  Or we choose to tell users upfront that we are not handling this?  I would prefer the functionality to cover broad cases if the efforts are reasonable. 

We don't want to find out these corner cases.  We should either implement them or documents them. 

It is fine for this PR not to implementation. But would be good to know the limitation and the plan.

lib/Transforms/InsertGpuAllocs.cpp
Jianhui-Li: Does this pass work with Vulkan runner?

lib/Transforms/InsertGpuAllocs.cpp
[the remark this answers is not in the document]
Jianhui-Li: I guess this is to find the use of memref  (produced or consumed) within the device kernel, whose memory buffers are likely being prepared by outside the device kernel. That's why it looks for load/store. 

For call, the parameter may or may not be defined outside the device kernel. I guess they are collected and then verified later.

lib/Transforms/InsertGpuAllocs.cpp
```diff
@@ -0,0 +1,363 @@
...
+
+            if (memInterface.hasEffect<mlir::MemoryEffects::Write>())
+              (onDevice ? ret.deviceWrite : ret.hostWrite) = true;
+
+            continue;
+          }
+          if (mlir::isa<mlir::func::CallOp>(user)) {
+            bool onDevice = user->getParentOfType<mlir::gpu::LaunchOp>();
+            (onDevice ? ret.deviceRead : ret.hostRead) = true;
+            (onDevice ? ret.deviceWrite : ret.hostWrite) = true;
```
Jianhui-Li, line 189: What is the use for "deviceRead " and "deviceWrite"? Do the code do any optimization for these tags?

lib/Transforms/InsertGpuAllocs.cpp
```diff
@@ -0,0 +1,363 @@
...
+
+      auto loc = getGlobalOp.getLoc();
+      auto access = it.second;
+      builder.setInsertionPointAfter(getGlobalOp);
+      auto getGlobalOpResult = getGlobalOp.getResult();
+      auto memrefType = getGlobalOp.getType().cast<mlir::MemRefType>();
+      auto rank = static_cast<unsigned>(memrefType.getRank());
+      filter.clear();
+      dims.clear();
+      for (auto i : llvm::seq(0u, rank)) {
```
Jianhui-Li, line 235: Does this cover shape with unknown rank, unkown dim, and known rank/dim? Do you have test cases for those?

Jianhui-Li (changes requested):

docs/Transforms/InsertGpuAllocs.md
Jianhui-Li: Are there any ways to remove these lines? 
  %3 = memref.cast %0 : memref<8xf32> to memref<?xf32>
  %4 = memref.cast %1 : memref<8xf32> to memref<?xf32>
  %5 = memref.cast %2 : memref<8xf32> to memref<?xf32>
  call @fillResource1DFloat(%3, %cst_0) : (memref<?xf32>, f32) -> ()
  call @fillResource1DFloat(%4, %cst) : (memref<?xf32>, f32) -> ()
  call @fillResource1DFloat(%5, %cst_1) : (memref<?xf32>, f32) -> ()

lib/Transforms/InsertGpuAllocs.cpp
nbpatel: added a limitations section in the doc description

lib/Transforms/InsertGpuAllocs.cpp
nbpatel: no, vulkan runnner has different set of passes that it runs and we haven't tried this pass with vulkan...

docs/Transforms/InsertGpuAllocs.md
nbpatel: done

docs/Transforms/InsertGpuAllocs.md
nbpatel: removed

lib/Transforms/InsertGpuAllocs.cpp
nbpatel: added more comments in the code

test/Transforms/set-spirv-capability.mlir
nbpatel: changed

lib/Transforms/InsertGpuAllocs.cpp
nbpatel: it works with known rank and unknown dims and obviously with static shapes ...not other cases..I will add a test case for dynamic dim case

include/imex/Transforms/Transforms.h
```diff
@@ -0,0 +1,52 @@
...
+#include <mlir/Dialect/MemRef/IR/MemRef.h>
+#include <mlir/Dialect/SCF/IR/SCF.h>
+#include <mlir/Dialect/SPIRV/IR/SPIRVDialect.h>
+#include <mlir/Dialect/SPIRV/IR/SPIRVOps.h>
+#include <mlir/Dialect/SPIRV/IR/TargetAndABI.h>
+#include <mlir/Dialect/SPIRV/Transforms/SPIRVConversion.h>
+#include <mlir/Pass/Pass.h>
+#include <mlir/Target/SPIRV/Serialization.h>
+#include <mlir/Transforms/DialectConversion.h>
+#include <mlir/Transforms/GreedyPatternRewriteDriver.h>
```
drprajap, line 40: it is better to move all these includes in cpp file to limit the inclusion scope..

lib/Transforms/InsertGpuAllocs.cpp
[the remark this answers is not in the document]
Hardcode84: This pass was specifically written to be run before kernel outlining, it is much harder to work on outlined kernel and generally not worth it.

lib/Transforms/InsertGpuAllocs.cpp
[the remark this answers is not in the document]
Hardcode84: This pass should be run after all linalg/affine ops were translated to scf, supporting them is too much work without any actual benefits.

lib/Transforms/InsertGpuAllocs.cpp
```diff
@@ -0,0 +1,383 @@
...
+              // Limitation is that this pass needs to be be run before the
+              // kernel outlining since kernel outlinging with convert the
+              // gpu.launch OP to gpu.launch_func.
+              if (!op->getParentOfType<mlir::gpu::LaunchOp>())
+                return mlir::WalkResult::advance();
+
+              if (!isMemReadWriteOp(op))
+                return mlir::WalkResult::advance();
+
+              auto memref = getMemReadWriteOp(op);
```
Jianhui-Li, line 125: getMemReadWriteOp() sounds like returning an OP not memref.

Jianhui-Li (approved):

PR #311, opened 2022-09-05
Add env region canonicalizations

Hardcode84:
Canonicalizations:
* Remove empty env region op (already covered by `NoSideEffect` handling, added test)
* Propagate yield values, defined outside region
* Merge nested regions with same environment and args
* Remove duplicated and unused yield args
* Merge adjacent regions with same environment and args


fschlimb (commented):
Thanks for including me, this is interesting.
However, this needs some description of the functional flow of each pass in oder for me to do a meaningful PR.

Hardcode84:
Added some comments

PR #314, opened 2022-09-08
PtensorType

fschlimb:
* Bringing the `PTensorType` closer to what's defined in the RFC.
  * Getting lowered to `tuple<rtensor, device, team, handle>`
  * Expanded to multiple arguments on function boundaries 
  * added `init_ptensor` and `extract_rtensor` ops
* More closely following upstream formatting (camelBack).
* Improved docs/comments

Without going down to a much more sophisticated analysis, I currently do not know how to make this work if `device`/`team`/`handle` types are not statically known. In other words, I do not know how to make this for `AnyType`. The reason is that the type-converter is missing the necessary context, e.g. it cannot determine the type of these unless it's encoded in the `PTensorType`. I tried to use types as type-parameters but I got stuck because MLIR cannot cope with `NonType` types in `TypeRanges` etc.. Without a default type which signals non-existence (of device etc) the whole idea of having them optional does not work, though. @Hardcode84 or anyone else, any suggestions?

Notice that `device`, `team` and `handle` are not always statically known, so they cannot be parameters of the `PTensorType`.

Also, more general, I think it is not a good idea to have these parameters of arbitrary type or even optional on function boundaries: it will disallow a simple and generic way of calling a function accepting or returning a ptensor. This is because the function signature depends on the potentially dynamically assigned device/distributed attributes.

Said all this I tend to think we should go back to having fixed types for representing devices, teams and handles.

Right now `device`, `team` and `handle` are represented with dummy `i1` types.

I'll make the tests more meaningful once we concluded the above problem. Notice that the imex-runner test actually runs a mlir-program using ptensor dialect and produces the right result.


fschlimb:
Sorry, this was using a git-consfused branch. I opened a new PR fixing this.

PR #315, opened 2022-09-08
Ptensor/ptensortype

fschlimb:
* Bringing the `PTensorType` closer to what's defined in the RFC.
  * Getting lowered to `tuple<rtensor, device, team, handle>`
  * Expanded to multiple arguments on function boundaries 
  * added `init_ptensor` and `extract_rtensor` ops
* More closely following upstream formatting (camelBack).
* Improved docs/comments

Without going down to a much more sophisticated analysis, I currently do not know how to make this work if `device`/`team`/`handle` types are not statically known. In other words, I do not know how to make this for `AnyType`. The reason is that the type-converter is missing the necessary context, e.g. it cannot determine the type of these unless it's encoded in the `PTensorType`. I tried to use types as type-parameters but I got stuck because MLIR cannot cope with `NonType` types in `TypeRanges` etc.. Without a default type which signals non-existence (of device etc) the whole idea of having them optional does not work, though. @Hardcode84 or anyone else, any suggestions?

Notice that `device`, `team` and `handle` are not always statically known, so they cannot be parameters of the `PTensorType`.

Also, more general, I think it is not a good idea to have these parameters of arbitrary type or even optional on function boundaries: it will disallow a simple and generic way of calling a function accepting or returning a ptensor. This is because the function signature depends on the potentially dynamically assigned device/distributed attributes.

Said all this I tend to think we should go back to having fixed types for representing devices, teams and handles.

Right now `device`, `team` and `handle` are represented with dummy `i1` types.

I'll make the tests more meaningful once we concluded the above problem. Notice that the imex-runner test actually runs a mlir-program using ptensor dialect and produces the right result.


Hardcode84:
Few immediate comments:
* Can we make it `ptensor.ptensor<?xi64>` instead of `ptensor.ptensor<tensor<?xi64>>`, I don't see much value in repeating `tensor` again inside the type
* bool `onDevice` is definitely not enough, we need attribute, which holds all device specific info. Also, as it was discussed previously `dist` should be part of the attribute and not part of pternsor type itself
* I don't see why we need op specifically for arange, it is better just to have some generic op, which have op type as attribute
* I would prefer for ptersor dialect to only concentrate on high level numpy stuff, most of its potential users are not interested in distributed stuff (including me at this point), and all distribution related things should live in theirs own dialect

fschlimb:
@Hardcode84 thanks for your comments.

> * Can we make it `ptensor.ptensor<?xi64>` instead of `ptensor.ptensor<tensor<?xi64>>`, I don't see much value in repeating `tensor` again inside the type

True, there is probably no value for the syntax. It keeps the implementation lean, though. Do you know about a declarative way to get this done? Or any other way to get this done without re-write all the tensor-parsing/printing (dimensionality/shape, eltype)?

> * bool `onDevice` is definitely not enough, we need attribute, which holds all device specific info. Also, as it was discussed previously `dist` should be part of the attribute and not part of pternsor type itself

Please read my comments above. `device`/`team` are a dynamic attributes, they cannot be a parameter of the type. `dist` and `onDevice` are type-parameters which allow static code optimizations by enabling/disabling code generation for offload/distribution. The dynamic values for `device`/`team` are passed into the creation ops.

> * I don't see why we need op specifically for arange, it is better just to have some generic op, which have op type as attribute

We had discussions about this and the outcome is the reflected in the RFC. The reasoning is in the document. Please provide technical arguments if you want to re-open the discussion by explaining in which respect it is better and how.

> * I would prefer for ptensor dialect to only concentrate on high level numpy stuff, most of its potential users are not interested in distributed stuff (including me at this point), and all distribution related things should live in theirs own dialect

By default use(r)s of the dialect don't see any of `device` or `team` (except on the function boundary, see my comments above). The lowering pass of course has to handle some offload/distribution stuff (which is by default disabled, see above). Again, the RFC explains the rational behind having this in one pass; if you want to re-open the discussion please provide technical arguments ("should be something" is not a technical argument since this is all about making tradeoffs). Suggestions for better handling at the function boundaries are definitely highly appreciated.

fschlimb:
The latest commits 
*  separate concerns of distribution and tensor math; we now have 3 passes:
   * --ptensor-dist
   * --convert-dist-to-standard
   * --convert-ptensor-to-linalg
* update RFC to account for the above and to clarify ptensor type and passes
* make tests more meaningful

@diptorupd any comments?

fschlimb:
@chudur-budur

diptorupd:
@fschlimb Sorry for chiming in late. 

Firstly, a small nit. I am not quite in favour of the `team` name. How about `distribution`?

> Please read my comments above. `device`/`team` are a dynamic attributes, they cannot be a parameter of the type. `dist` and `onDevice` are type-parameters which allow static code optimizations by enabling/disabling code generation for offload/distribution. The dynamic values for `device`/`team` are passed into the creation ops.

I statement about `device` and `team`as dynamic attributes is not clear. Are you saying that these attributes can change for a given ptensor object?

For data-flow analysis, the `device` and `team` attributes should be statically ascertainable. I propose making the "filter string" statically part of the type. If there is a need, we can allow the `device` to be a `?` for a ptensor with unknown device type. 

The static information about the device will make it possible to validate compute follows data rules statically, rather than defer them to runtime. We will also be able to apply data flow analysis before any kind of fusion or DCE, CSE conversions.

Tying the device attribute to the filter string may seem SYCL specific, but in implementation it can be kept generic and can be left to the underlying runtime to reason about. We only need it to be a constant at compile time to aid static analysis of the code.

fschlimb:
> @fschlimb Sorry for chiming in late.
> 
> Firstly, a small nit. I am not quite in favour of the `team` name. How about `distribution`?

Distribution does not make sense to me. Team is the equivalent of device. Dist(ribution) is the equivalent of "offload".

> I statement about `device` and `team`as dynamic attributes is not clear. Are you saying that these attributes can change for a given ptensor object?

No. I was trying to allow device/team to be known at runtime only (but ssa). Apparently, they need to be constant to allow optimizations like fusion. If we make device/team attributes of PTensor, then it does not allow computing the device within the compiled code. For example, a runtime library might want to assign certain tasks to specific devices. 

I agree that determining the device/team "dynamically" on the fly will not be the common case. Still, I wanted to be able to support it.

> For data-flow analysis, the `device` and `team` attributes should be statically ascertainable. I propose making the "filter string" statically part of the type. If there is a need, we can allow the `device` to be a `?` for a ptensor with unknown device type.

Ok, with a dedicated identifier like '?' (even though it is always better to avoid special values) it seems we could support the decision about the device at runtime. It would be somewhat simpler to explain/describe. I don't know if/how it is possible to have a string as a type-attribute. I'll give it some thought.

> The static information about the device will make it possible to validate compute follows data rules statically, rather than defer them to runtime. We will also be able to apply data flow analysis before any kind of fusion or DCE, CSE conversions.

Yes, but this does not mean that the device needs to be a constant attribute of the type. As long as the device is provided as a constant value all the above should be possible - no matter if it is a type attribute or a constant parameter to an op.

> Tying the device attribute to the filter string may seem SYCL specific, but in implementation it can be kept generic and can be left to the underlying runtime to reason about. We only need it to be a constant at compile time to aid static analysis of the code.

Agree.

fschlimb:
Ok I will merge this now and then issue  PR to update the SHA. Otherwise the SHA update will become a mess.

PR #318, opened 2022-09-09
Switch gpu lowering to env region

Hardcode84:
* Introduce `GPURegionDescAttr` for env region (`device` field is not yet used)
* Add additional builder to `EnvironmentRegionOp`
* Switch gpu lowering from `GpuRange` attribute to env region


Hardcode84:
@nbpatel I've changed `ParallelLoopGPUMappingPass` to use env region but switched L0 runner to upstream `createGpuMapParallelLoopsPass` so GPU tests are not affected

PR #321, opened 2022-09-11
Introduce new Execution engine and new Numba interop

Hardcode84:
* Add `ExecutionEngine` to imex core, existing mlir one is not suitable for us:
  * Doesn't support dynamically adding and removing modules
  * Contains 'packed' stuff, which we don't need and which cannot be disabled
* Instead passing LLVM bitcode to Numba side, pass function pointer to compiled code
* Move `RetainOp` lowering back to `lower_to_llvm.cpp` as it contains Numba-specific lowering
* Define our own versions of `NRT_Incref`m `NRT_Decref` and `NRT_MemInfo_data_fast` to reduce dependency on Numba runtime
* Compiled modules are not being freed at the moment and they will leak, but they are not freed in baseline numba as well, so no regression here

### DIFF
--- a/mlir/lib/ExecutionEngine/execution_engine.cpp
+++ b/mlir/lib/ExecutionEngine/execution_engine.cpp
@@ -154,27 +154,6 @@ imex::ExecutionEngine::ExecutionEngine(ExecutionEngineOptions options)
       objectLayer->setAutoClaimResponsibilityForObjectSymbols(true);
     }
 
-    // Resolve symbols from shared libraries.
-    //    for (auto libPath : options.sharedLibPaths) {
-    //      auto mb = llvm::MemoryBuffer::getFile(libPath);
-    //      if (!mb) {
-    //        llvm::errs() << "Failed to create MemoryBuffer for: " << libPath
-    //               << "\nError: " << mb.getError().message() << "\n";
-    //        continue;
-    //      }
-    //      auto &jd = session.createBareJITDylib(std::string(libPath));
-    //      auto loaded = DynamicLibrarySearchGenerator::Load(
-    //          libPath.data(), dataLayout.getGlobalPrefix());
-    //      if (!loaded) {
-    //        llvm::errs() << "Could not load " << libPath << ":\n  " <<
-    //        loaded.takeError()
-    //               << "\n";
-    //        continue;
-    //      }
-    //      jd.addGenerator(std::move(*loaded));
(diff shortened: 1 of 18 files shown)


mlir/lib/ExecutionEngine/execution_engine.cpp
```diff
@@ -0,0 +1,299 @@
...
+imex::ExecutionEngine::ExecutionEngine(ExecutionEngineOptions options)
+    : cache(options.enableObjectCache ? new SimpleObjectCache() : nullptr),
+      gdbListener(options.enableGDBNotificationListener
+                      ? llvm::JITEventListener::createGDBRegistrationListener()
+                      : nullptr),
+      perfListener(nullptr) {
+  if (options.enablePerfNotificationListener) {
+    if (auto *listener = llvm::JITEventListener::createPerfJITEventListener())
+      perfListener = listener;
+    else if (auto *listener =
```
AlexanderKalistratov, line 129: And if createIntelJITEventListener returns nullptr too?

mlir/lib/ExecutionEngine/execution_engine.cpp
```diff
@@ -0,0 +1,299 @@
...
+
+    // COFF format binaries (Windows) need special handling to deal with
+    // exported symbol visibility.
+    // cf llvm/lib/ExecutionEngine/Orc/LLJIT.cpp LLJIT::createObjectLinkingLayer
+    if (targetTriple.isOSBinFormatCOFF()) {
+      objectLayer->setOverrideObjectFlagsWithResponsibilityFlags(true);
+      objectLayer->setAutoClaimResponsibilityForObjectSymbols(true);
+    }
+
+    // Resolve symbols from shared libraries.
```
AlexanderKalistratov, line 157: Commented code

mlir/lib/ExecutionEngine/execution_engine.cpp
Hardcode84: This was directly copypasted from mlir examples. It both functions fail, execution engine should still work, just without debugging support.

mlir/lib/ExecutionEngine/execution_engine.cpp
Hardcode84: Yeah, this was just copypasted from examples, but I dont need this as the moment so I'll probably just remove this.

PR #323, opened 2022-09-12
Add RFC for GPUX Dialect

nbpatel:
This PR adds the RFC for the GPUX dialect. GPUX dialect is an extension of the upstream GPU dialect with few ops extended from the upstream GPU dialect to take stream as an argument. More details in the RFC.


fschlimb:
Could you please add a meaningful message instead of the standard task-list?

Is the idea to destroy every stream which has been created in the compilation unit just before exiting? How do you detect when it is safe to destroy?

With #283 we assume we will have a region which defines the location (e.g. device/stream) of an array/tensor and operations in it. This is needed to allow compute-follows-data semantics. Such a region can easily be used for everything described here. However, finding the right points for inserting Create/DestroyStreams in compute-follows-data cases is complex unless we have some marker like the suggested region.

Can we use a region as suggested in #283 instead of of explicit Create/DestroyStream? The AddIntelGpuStream pass can then easily insert CreateStream at the beginning of the region and a DestroyStream at its end.

docs/rfcs/IntelGpuDialect.md
```diff
@@ -0,0 +1,122 @@
+# RFC: Intel GPU Dialect (I-GPU)
+
+Nishant Patel
+
+## Summary
+
+We propose the Intel Gpu Dialect as a custom Intel dialect to unlock the underlying runtimes (Level zero & Sycl) best suited for Intel Gpus.
+This Dialect will be at a lower level than the upstream GPU Dialect.
```
silee2, line 8: I think it should be pointed out that 
1) Intel GPU dialect is a conversion target for upstream GPU dialect as an implementation dialect.
2) Goal is not to provide a lower level alternative to all GPU dialect ops and provide just things lacking in upstream GPU dialect.
But on the other hand may be too much information for a summary.

silee2:
> Could you please add a meaningful message instead of the standard task-list?
Agree. PRs should have a short description. For this PR the summary section of documentation could be copied over.

silee2:
> Is the idea to destroy every stream which has been created in the compilation unit just before exiting? How do you detect when it is safe to destroy?

We don't have any kernel lifetime analysis for now. So streams would be destroyed at the end of the caller function calling gpu.launch_op or right after the gpu.launch_op.
@nbpatel Can you clarify?

docs/rfcs/IntelGpuDialect.md
```diff
@@ -0,0 +1,122 @@
...
+    %c1 = arith.constant 1 : index
+    %cst = arith.constant 2.200000e+00 : f32
+    %cst_0 = arith.constant 1.100000e+00 : f32
+    %cst_1 = arith.constant 0.000000e+00 : f32
+    %0 = "intel_gpu.create_stream"() : () -> !intel_gpu.OpaqueType
+    .
+    .
+    .
+    .
+    %6 = gpu.launch_func async [%5] @main_kernel::@main_kernel blocks in (%c8, %c1, %c1) threads in (%c1, %c1, %c1) args(%memref : memref<8xf32>, %memref_2 : memref<8xf32>, %memref_4 : memref<8xf32>)
```
Jianhui-Li, line 66: Don't user need to specify which stream to use when launch a kernel function?

docs/rfcs/IntelGpuDialect.md
```diff
@@ -0,0 +1,122 @@
...
+    %9 = llvm.mlir.constant(64 : i64) : i64
+    %10 = llvm.call @iGpuMemAlloc(%8, %9, %3) : (i64, i64, !llvm.ptr<i8>) -> !llvm.ptr<i8>
+    .
+    .
+    .
+    %62 = llvm.call @iGpuModuleLoad(%3, %60, %61) : (!llvm.ptr<i8>, !llvm.ptr<i8>, i64) -> !llvm.ptr<i8>
+    .
+    %66 = llvm.call @iGpuKernelGet(%3, %62, %65) : (!llvm.ptr<i8>, !llvm.ptr<i8>, !llvm.ptr<i8>) -> !llvm.ptr<i8>
+    .
+    llvm.call @iGpuLaunchKernel(%3, %66, %0, %1, %1, %1, %1, %1, %67, %108, %109) : (!llvm.ptr<i8>, !llvm.ptr<i8>, i64, i64, i64, i64, i64, i64, i32, !llvm.ptr<struct<(ptr<i8>, i64)>>, !llvm.ptr<ptr<i8>>) -> ()
```
Jianhui-Li, line 103: If user creates multiple streams, how the lowering decides which stream to use?  I think user should pass the stream to launch kernel function.

docs/rfcs/IntelGpuDialect.md
```diff
@@ -0,0 +1,122 @@
...
+```
+
+## Features to be added in subsequent PR's (either in the dialect or via a pass operating on this dialect):
+1. More ops can be added to dialect going forward based on the use cases and requirements.
+2. Caching of the module creation needs to be added via a pass working on this dialect.
+3. Currenlty, the stream creation and destruction happens for every function. A pass to manage stream will be added where the stream is created only once at the beginning of the program and used throughout.
+
+## Alternative
+
+The alternative approch here is to add this operation in the upstream dialect and extend the upstream GPU dialect. But, we feel that having our own dialect will give us more command and control over how we want to design it.
```
Jianhui-Li, line 118: I would expect the upstream GPU would evolve to expose stream support to GPU dialect user.

docs/rfcs/IntelGpuDialect.md
Jianhui-Li: From the description, it is very clear what proposed is not a full set of dialect, just an extension to upstream GPU dialect.    If possible, we would like to upstream it.

Jianhui-Li (changes requested):

docs/rfcs/IntelGpuDialect.md
nbpatel: yes, if user specify the stream then we just use that, or else we will create a default one and launch kernel...

docs/rfcs/IntelGpuDialect.md
nbpatel: the launch kernel takes stream as an input when its lowered to llvm IR..
```llvm.call @iGpuLaunchKernel(%3, %66, %0, %1, %1, %1, %1, %1, %67, %108, %109)```

here %3 is the stream/queue

**Comment by nbpatel** [comment]
> > Is the idea to destroy every stream which has been created in the compilation unit just before exiting? How do you detect when it is safe to destroy?
> 
> We don't have any kernel lifetime analysis for now. So streams would be destroyed at the end of the caller function calling gpu.launch_op or right after the gpu.launch_op. @nbpatel Can you clarify?

Yes, stream management is very naive to start with, the stream is destroyed at the end of every function,  but in subsequent PR's we will make it more flexible, i.e user providing stream, same stream usage across multiple functions, multi stream management.. etc

File: docs/rfcs/IntelGpuDialect.md (hunk 3 above)
Comment (fschlimb) [reply docs/rfcs/IntelGpuDialect.md:103]: How does the lowering decide which stream to use?

**Comment by fschlimb** [comment]
> The create/destroy and using stream for individual kernel launch is fine level control than the region idea. A region using one stream/context can be supported by this programming model. There are other use cases like the user like to use different stream to launch different kernel.

I don't understand. You can have multiple regions as you can have multiple create-streams. Where is the difference in granularity? Where would the region reside?

File: docs/rfcs/IntelGpuDialect.md (hunk 3 above)
Comment (Jianhui-Li) [reply docs/rfcs/IntelGpuDialect.md:103]: This RFC needs to be changed. User decide which stream to use for each kernel, not compiler.

File: docs/rfcs/IntelGpuDialect.md (hunk 4 above)
Comment (Jianhui-Li) [reply docs/rfcs/IntelGpuDialect.md:118]: > > The create/destroy and using stream for individual kernel launch is fine level control than the region idea. A region using one stream/context can be supported by this programming model. There are other use cases like the user like to use different stream to launch different kernel.
> 
> I don't understand. You can have multiple regions as you can have multiple create-streams. Where is the difference in granularity? Where would the region reside?

Please note that this PRC needs to be changed to "User decide which stream to use for each kernel, not the compiler".  This give user fine-level control.  The region may also work but region usually contain multiple kernels dispatching to one same stream.

File: docs/rfcs/GPUXDialect.md [hunk 5]
```diff
@@ -0,0 +1,226 @@
...
+
+The wait op will need to be provided with the stream to wait on. This op is an extension of upstream gpu.wait op with one added argument for stream. Stream here will be an optional argument, so when upstreaming the code, the upstream GPU dialect users have no change in their usage of this op.
+
+## Example Usage in case of user provided stream:
+
+// IR before our custom transformation pass:
+
+```
+
+func.func @main(%stream : !gpux.StreamType) attributes {llvm.emit_c_interface} {
```
Jianhui-Li, line 112: This IR doesn't make sense. When user transforms to GPU dialect, user either choose to not specify stream. If user choose to user stream, it is user's responsibility to use the stream to GPU ops.

docs/rfcs/GPUXDialect.md
```diff
@@ -0,0 +1,226 @@
...
+  %c8 = arith.constant 8 : index
+  %c1 = arith.constant 1 : index
+  %cst = arith.constant 2.200000e+00 : f32
+  %cst_0 = arith.constant 1.100000e+00 : f32
+  %cst_1 = arith.constant 0.000000e+00 : f32
+  %0 = gpux.alloc (%stream) : (!gpux.StreamType) -> memref<8xf32>
```
Jianhui-Li, line 145: It doesn't make sense. How the optimization pass would know that it needs tot take the stream passed by the function and assign it to this operations?  What about user pass two streams in the main function parameters? What if the operations within  program itself needs to two streams?  There should be a separate optimization pass to decide which stream to use, and that optimization pass need to decide which op to use which stream. Not just pass it as parameter, hoping another optimization pass to fix the rest.

docs/rfcs/GPUXDialect.md
```diff
@@ -0,0 +1,226 @@
...
+  return
+}
+
+```
+
+// IR after our custom transformation pass:
+
+```
+
+func.func @main() attributes {llvm.emit_c_interface} {
```
Jianhui-Li, line 194: This pass makes sense to me.  This means user doesn't want to control at all.

Jianhui-Li (changes requested):

diptorupd:
@nbpatel I still have reservations against the name. Level Zero and SYCL are not GPU only. With the right drivers we can target CPU or any other device as well.

Jianhui-Li:
> @nbpatel I still have reservations against the name. Level Zero and SYCL are not GPU only. With the right drivers we can target CPU or any other device as well.

The name "GPU" dialect is from upstream. It tries to abstract a common runtime function and some workgroup level functions for all GPUs.  This PR is about extending the upstream "GPU" dialect. We don't have a plan to expose Level Zero and SYCL runtime as a dialect in a short time.  The current plan is to implement the GPU dialect using level Zero and SYCL runtime.  

If the SYCL dialect intends to bring a high-level abstraction than GPU dialect, that would be a separate dialect, which is separate effort.  I guess the value would be that MLIR lowering passes don't need to separate host and device code, and covert the full computation to SYCL dialect, which allows some SYCL specific optimization.  THis PR doesn't intend to that.  

For L0, I don't think it worth to have dialect since it is a pure runtime.  Just the ability of running device kernel on CPU may not justify a dialect. I think a dialect needs to support certain compile-time optimization enabled by the runtime capability.  

 @diptorupd please explain the vision of SYCL/L0 dialect here in case I misunderstand.  To introduce a new dialect, we need to justify what kind of compile-time optimization we enabled by introducing that as a dialect.

diptorupd:
> The current plan is to implement the GPU dialect using level Zero and SYCL runtime.

@Jianhui-Li It might be a gap in my understanding, can you elaborate on this point? If our iGPU dialect is going to use L0 or SYCL, I can potentially use it to launch kernels on any SYCL or L0 device not just a GPU. 

> If the SYCL dialect intends to bring a high-level abstraction than GPU dialect, that would be a separate dialect, which is separate effort. 

Agreed and such a dialect will be higher-level than the GPU dialect and will potentially lower to the GPU dialect. Codeplay was working on such a dialect at some point at least as an R&D project.

Jianhui-Li:
> > The current plan is to implement the GPU dialect using level Zero and SYCL runtime.
> 
> @Jianhui-Li It might be a gap in my understanding, can you elaborate on this point? If our iGPU dialect is going to use L0 or SYCL, I can potentially use it to launch kernels on any SYCL or L0 device not just a GPU.
>
It means that the L0/SYCL runtime are only used to launch kernel to GPU since these kernels we are handling are GPU kernel.  If you want something like "XPU" dialect - underneath the kernel could be dispatch to CPU or GPU, it is out of the scope of this PR.

Jianhui-Li:
> > > The current plan is to implement the GPU dialect using level Zero and SYCL runtime.
> > 
> > 
> > @Jianhui-Li It might be a gap in my understanding, can you elaborate on this point? If our iGPU dialect is going to use L0 or SYCL, I can potentially use it to launch kernels on any SYCL or L0 device not just a GPU.
> 
> It means that the L0/SYCL runtime are only used to launch kernel to GPU since these kernels dispatched by GPU dispatch dialect is GPU kernel - it is defined as the name and so it is what the upper software layer expected.  If you want something like "XPU" dialect - underneath the kernel could be dispatch to CPU or GPU, it is out of the scope of this PR.

docs/rfcs/GPUXDialect.md
nbpatel: >>>it is user's responsibility to use the stream to GPU ops.

how will the user pass the stream to gpu ops ? they have to use this GPUX dialect to pass the stream to GPU ops...are you saying user will have their own pass to convert to the GPUX dialect which adds stream as an input to the upstream gpu ops?

nbpatel:
> @nbpatel I still have reservations against the name. Level Zero and SYCL are not GPU only. With the right drivers we can target CPU or any other device as well.

Our dialect is specifically targeted for GPU and is an extension of the upstream dialect, hence the name GPUX (where X stands for extension)...we do not say anywhere in the RFC that this dialect is generic for all backends using L0 and SYCL

docs/rfcs/GPUXDialect.md
fschlimb: @Jianhui-Li That's the part which do not understand. The mere existence of a stream-creation operation somewhere in the code give no hint about where to use it. Hence we suggested using regions (see #283). Can you explain how an analysis would be able to do this mapping? What's the contract for linking create_stream to the alloc?

docs/rfcs/GPUXDialect.md
fschlimb: Why not? From our perspective that's exactly what happens with compute-follows-data. There can be a default behavior, but it seems important to be able to control this.

diptorupd:
> This dialect is specifically targeted for GPU and is an extension of the upstream dialect, hence the name GPUX (where X stands for extension)...we do not say anywhere in the RFC that this dialect is generic for all backends using L0 and SYCL

I realize that. My question is: will it be possible for me as a consumer of GPUX to use the dialect for launching a kernel on a CPU device? Or are you saying it is not possible?

diptorupd:
> I realize that. My question is: will it be possible for me as a consumer of GPUX to use the dialect for launching a kernel on a CPU device? Or are you saying it is not possible?

More directly, what will GPUX do if I pass a CPU stream as an argument?

diptorupd:
@nbpatel quoting from your RFC

> The stream here is basically a SYCL/L0 queue, with default context and device. If the stream is user provided we assume the user is adhering to the Stream Type (!gpux.StreamType which is a OpaqueType pointer).

I can very easily change environment flags `SYCL_DEVICE_FILTER="cpu"` to change the default device to CPU. (refer: https://intel.github.io/llvm-docs/EnvironmentVariables.html#sycl_device_filter)

> gpux.create_stream (gpux::CreateStreamOp)

How can you be sure that the stream is a GPU stream? As the behavior is controlled by SYCL runtime or L0 runtime, I can get an CPU, FPGA stream if I want. What is preventing me from doing so?

diptorupd:
> It means that the L0/SYCL runtime are only used to launch kernel to GPU since these kernels we are handling are GPU kernel.

@Jianhui-Li What is a GPU kernel? I mean how is it represented? Is it already PTX or SPIRV at the point where it gets to GPUX?

diptorupd:
> @Jianhui-Li What is a GPU kernel? I mean how is it represented? Is it already PTX or SPIRV at the point where it gets to GPUX?

I read the doc for the upstream GPU dialect. Here is what constitutes a GPU kernel. Based on this, the next stage in the compilation pipeline is perhaps `GPU -> LLVM -> PTX`, and in our case `GPUX -> SPIR`? Is that a correct assumption?

```
module attributes {gpu.container_module} {

  // This module creates a separate compilation unit for the GPU compiler.
  gpu.module @kernels {
    func.func @kernel_1(%arg0 : f32, %arg1 : memref<?xf32, 1>)
        attributes { nvvm.kernel = true } {

      // Operations that produce block/thread IDs and dimensions are
      // injected when outlining the `gpu.launch` body to a function called
      // by `gpu.launch_func`.
      %tIdX = gpu.thread_id x
      %tIdY = gpu.thread_id y
      %tIdZ = gpu.thread_id z

      %bDimX = gpu.block_dim x
      %bDimY = gpu.block_dim y
      %bDimZ = gpu.block_dim z

      %bIdX = gpu.block_id x
      %bIdY = gpu.block_id y
      %bIdZ = gpu.block_id z

      %gDimX = gpu.grid_dim x
      %gDimY = gpu.grid_dim y
      %gDimZ = gpu.grid_dim z

      "some_op"(%bx, %tx) : (index, index) -> ()
      %42 = load %arg1[%bx] : memref<?xf32, 1>
    }
  }

  %t0 = gpu.wait async
  gpu.launch_func
      async                           // (Optional) Don't block host, return token.
      [%t0]                           // (Optional) Execute only after %t0 has completed.
      @kernels::@kernel_1             // Kernel function.
      blocks in (%cst, %cst, %cst)    // Grid size.
      threads in (%cst, %cst, %cst)   // Block size.
      dynamic_shared_memory_size %s   // (Optional) Amount of dynamic shared
                                      // memory to allocate for a workgroup.
      args(%arg0 : f32,               // (Optional) Kernel arguments.
           %arg1 : memref<?xf32, 1>)
}
```

Jianhui-Li:
>>I read the doc for the upstream GPU dialect. Here is what constitutes a GPU kernel. Based on this, the next stage in the compilation pipeline is perhaps GPU -> LLVM -> PTX, and in our case GPUX -> SPIR? Is that a correct assumption?

Yes

diptorupd:
@Jianhui-Li @nbpatel 

I am collecting all my feedback based on the current RFC doc and your responses to my earlier review comments:

**1.  GPUX does not target GPU devices only.**

The following assertion made by @nbpatel is incorrect. We may not claim to do so in the RFC, but being based on SYCL runtime GPUX will in fact support launching kernels on any SPIR-V supporting device (assuming the compilation pipeline that @Jianhui-Li affirmed in https://github.com/intel/mlir-extensions/pull/323#issuecomment-1256582224 is always followed. 

> This dialect is specifically targeted for GPU and is an extension of the upstream dialect, hence the name GPUX (where X stands for extension)...we do not say anywhere in the RFC that this dialect is generic for all backends using L0 and SYCL

Assuming the scenario in [Example Usage in case of default stream](https://github.com/intel/mlir-extensions/blob/nishant_rfc/docs/rfcs/GPUXDialect.md#example-usage-in-case-of-default-stream) and a pipeline that does `linalg` -> `gpu` -> `gpux` -> `spirv`, I can easily force the kernel to be launched on a L0 or OpenCL CPU device. All I really have to do is set the `SYCL_DEVICE_FILTER="cpu"` environment variable for dpcpp to create a CPU SQL queue by default. 

If I use the open source dpcpp built with CUDA or ROC support, I can actually do even worse things. By setting  `SYCL_DEVICE_FILTER="cuda"` I can cause GPUX to generate SPIR-V and pass it to dpcpp telling it to launch a CUDA kernel that will cause an exception and most likely a runtime crash.

That the GPUX name is misleading is only a minor issue. **The real thing is that the assumption that GPUX is only launching kernels on Intel L0 GPU devices is incorrect.**

**2.  DPCPP has no concept of default stream**

I do not see the justification for the [Example Usage in case of default stream](https://github.com/intel/mlir-extensions/blob/nishant_rfc/docs/rfcs/GPUXDialect.md#example-usage-in-case-of-default-stream). As I pointed out in (1) it can lead to incorrect results. 

Moreover on a pedantic point about the following line in the motivation section of the RFC:

>The stream is a default stream if its not provided by user or it can be a user provided stream. The stream here is basically a SYCL/L0 queue, with default context and device

In DPCPP there is a notion of a default device, the one returned by the `default_selector`, and there is a notion of a default context, but there is no notion of a default queue for a device.

**3.  Handling async**

I do not follow the semantics of the `gpux.wait (gpux::WaitOp)`. 

> Stream here will be an optional argument, so when upstreaming the code, the upstream GPU dialect users have no change in their usage of this op.

How can the stream be an optional argument?

Going back to point (2), DPCPP and SYCL have no CUDA like notion of a default stream. If we call `gpux.wait()` with no stream what is the expected outcome? Is the assumption that we will get the "default" queue from the SYCL runtime and wait on it?
If so, it is not going to work. Each implicitly created queue by GPUX is separate and independent and once the reference to a queue is gone, you cannot wait on it.

**4.  [Example Usage in case of user provided stream](https://github.com/intel/mlir-extensions/blob/nishant_rfc/docs/rfcs/GPUXDialect.md#example-usage-in-case-of-user-provided-stream)**

The section [Example Usage in case of user provided stream](https://github.com/intel/mlir-extensions/blob/nishant_rfc/docs/rfcs/GPUXDialect.md#example-usage-in-case-of-user-provided-stream) needs lot more elucidation. 

Are users of GPUX dialects supposed to generate this function?

Let me provide an example taking the following snippet from the PTensor RFC, and ask specific questions:

`tensor.add(%pt1, %pt2): (tensor<...>, tensor<...>) -> tensor.tensor<...>`

Assuming the following pipeline:

`ptensor` -> `linalg` -> `gpu`

The outcome of the pipeline is something like (pardon my nascent MLIR language familiarity):

```
module {
    gpu.func @add_kernel(%A: memref<1024xf32>, %B: memref<1024xf32>, %C: memref<1024xf32>) {
        # vector add ...
        return
    }

    func.func @some_func() {
        %A = memref.alloc() : memref<1024xf32>
        %B = memref.alloc() : memref<1024xf32>
        %C = memref.alloc() : memref<1024xf32>
        
        %cf1 = constant 1.0 : f32
        
        linalg.fill(%A, %cf1) : memref<1024xf32>, f32
        linalg.fill(%B, %cf1) : memref<1024xf32>, f32
        linalg.fill(%C, %cf1) : memref<1024xf32>, f32
        
        call @add_kernel(%A, %B, %C) : (memref<1024xf32>, memref<1024xf32>, memref<1024xf32>) -> ()
        return
    }
}
```
Now the important question:

** Does GPUX require an user (like Ptensor) to change the signature of the function and pass in a stream?** 

That is the part I do not follow. The `func.func` at this stage is tied to some front end code and should bear close resemblance to that code, changing the signature makes the IR harder to follow. (I you do not believe me write a C function that returns a struct of three fp64 and see what LLVM IR gets generated:) )

** Or is the requirement that Ptensor should outline each possible GPUX region into a function and pass in the stream to that outlined function? That will be like the OpenMP outlining that Clang does.**

If that is the intention, even that is sub-optimal as the front-end now has to do more work to use GPUX. Either each front end has to write a custom GPU+GPUX lowerer or else generate the "dirty" IR and the run the GPUToGPUX pass.

I call it dirty IR because, we have a function that takes an unused argument and what if something like a `DeadArgumentEliminationPass` runs and cleans it up?

**5.  Memcpy and Memset**

We definitely need memcpy and memset in GPUX.

diptorupd:
The (4) point above is the main issue I have with the RFC.

@Hardcode84, I, and @fschlimb have discussed it at length and had an alternative proposal in the ptensor RFC. Ivan has already prototyped it in main.

Instead of an outlined function or changing the function signature of a user-level function it is much simpler (and elegant) for a front-end to add a [region op](https://github.com/intel/mlir-extensions/tree/refactor/docs/rfcs/20220804-ptensor#intel_syclgpu-region).

```
gpux.device_region('level_zero:gpu:1') {
  tensor.add(%pt1, %pt2): (tensor<...>, tensor<...>) -> tensor.tensor<...>
}
```
The region op is generated by the user (ptensor or anything higher than that) and persists through code-generation. Then when the GPUX transformation runs, it runs a region-level transformation to change whatever it needs to do. 

Refer: https://github.com/intel/mlir-extensions/pull/318

Hardcode84:
Some clarifications from current main/python pipeline
* It is not limited to GPU and potentially can run on any L0 device capable to consume spirv, there is currently hardcoded piece of code which always select first gpu device, but it is going to change after switch to CFD.
* Currently we are using both dev region and stream code on different stages, as dev region is more high-level concept and explicit stream management is useful for low level optimizations (e.g. spirv loading outlining). Current pipeline is `region` -> `gpu` -> `gpux` (`gpu_runtime` in main).
* Default stream is not supported, any ops with null stream or unconverted gpu ops won't be lowered and will cause compilation error.
* Currently, stream only created/destroyed on function entry/exit and never crosses function boundary, it is enough for kernel style API implementation, we may need a more advanced scheme for CFD but this is TBD at the moment.

diptorupd:
> It is not limited to GPU and potentially can run on any L0 device capable to consume spirv, there is currently hardcoded piece of code which always select first gpu device, but it is going to change after switch to CFD.

Exactly, ergo my point about the claim that GPUX supports only Intel GPUs being incorrect. 

My concern in (1) of  https://github.com/intel/mlir-extensions/pull/323#issuecomment-1256791895 is still not solved. Basically, there is no way to derive the code generation target without knowing the stream type at compile time. If we keep it as an opaque pointer and always generate SPIR-V we can end up with lots of unintended issues like the one I mentioned (sending a SPIRV kernel to a CUDA stream). For this reason, dpcpp does the SYCL kernel to target bitcode generation using JIT compilation.

One very quickly thought of hack that I and @fschlimb had come up was always use a well defined filter string in the `device_region` op. Unfortunately, that is not going to work. Just the string "gpu" or even "0" is a valid filter string and we are back to square one, moreover we will need to pass the stream object around as you said once we have more advanced programming model such as compute follows data (CFD).

We most likely need additional ops that will let us extract the platform from a queue/stream. The bitcode target (SPIRV, GCN, PTX) can be predicated on that. The more I think of it we really need a JIT to solve the GPUX to backend issue.

I am not necessarily saying the problem should be solved in GPUX design right away, but we have to be very clear as to what the dialect does and does not do. Especially, if our goal is to upstream it in some form.

I have also initiated a discussion with the compiler team working on an MLIR-SYCL dialect to see what their approach to the problem is.

Hardcode84:
> My concern in (1) of https://github.com/intel/mlir-extensions/pull/323#issuecomment-1256791895 is still not solved. Basically, there is no way to derive the code generation target without knowing the stream type at compile time.

In my mental model I've always assumed that device (or filter string from which device can be inferred) is always known statically to compilation pipeline. E.g. when we are passing dpctl arrays to `njit` function we will extract device (or filter string) from it and make it part of function types in IR. If we pass array with different device we just retrigger jit for the entire function.

diptorupd:
> always known statically to compilation pipeline.

Sure, but that is our specific JIT based use case. I think GPUX needs to be more general purpose. 

@nbpatel one thing missing from the RFC is usage scenarios. How upstream dialects like plier, ptensor should use the dialect. And how the downstream generation from GPUX o spir etc. is going to be handled.

fschlimb:
> To answer your question: **Or is the requirement that Ptensor should outline each possible GPUX region into a function and pass in the stream to that outlined function? That will be like the OpenMP outlining that Clang does.**
> 
> > Yes.

@nbpatel @Jianhui-Li Could you please explain what happens to such a function? How does the stream argument get used?

docs/rfcs/GPUXDialect.md
```diff
@@ -0,0 +1,176 @@
...
+
+Syntax:
+
+```
+operation ::= gpux.launch_func custom<AsyncDependencies>(type($asyncToken), $asyncDependencies)
+              $kernel
+              blocks in  ($gridSizeX, $gridSizeY, $gridSizeZ)
+              threads in ($blockSizeX, $blockSizeY, $blockSizeZ)
+              (dynamic_shared_memory_size $dynamicSharedMemorySize^)?
+              custom<LaunchFuncOperands>($gpux_stream, $operands, type($operands)) attr-dict
```
Jianhui-Li, line 87: Why $gpux_stream is part of device function parameter? It should be part of GPUX.launch_func(), so becomes the host code interacting with runtime, not the device kernel.

Jianhui-Li: What is the reason we don't support gpu.launch also?

nbpatel: yes, thats a mistake. it has to be next to kernel. will change it

We dont support gpu.launch is because we assume the user to run the upstream GpuKernelOutliningPass which would translate the gpu.launch to gpu.launch_func and outline the device kernel. Most of the lower passes expect the kernel to be outlined

Jianhui-Li:
Looks good to me.

docs/rfcs/GPUXDialect.md
```diff
@@ -0,0 +1,233 @@
+# RFC: GPUX Dialect (Extension of upstream GPU Dialect)
+
+Core MLIR Team
+
+## Summary
+
+We propose the GPUX dialect as an extension of the upstream GPU dialect. The GPUX dialect allows us to expose stream creation/destruction ops.These ops are required by the underlying runtimes (Level zero & Sycl) for explicit stream/context/device creation.
```
silee2, line 7: I think you can switching the order of the statements may look more natural. You can first mention some runtimes require explicit stream context and device since that is the reason for introducing GPUX dialect. And then move onto introducing new ops (stream creation/destruction) and op changes.

docs/rfcs/GPUXDialect.md
```diff
@@ -0,0 +1,233 @@
...
+
+## Motivation
+
+Upstream MLIR has a Dialect called the [GPU Dialect](https://mlir.llvm.org/docs/Dialects/GPU/) which provides middle-level abstractions for launching GPU kernels following a programming model similar to that of CUDA or OpenCL. The upstream dialect exposes all the operation required to launch the kernel on a GPU device. But, the drawback with the upstream dialect is it does not expose Stream as an operation in the dialect. For example, if the user wants to launch the kernel on a particular stream(provided by the user), there is no way to do that today. For the upstream dialect, stream is created internally while lowering to LLVM runtime calls and it refers specifically to the CUDA stream. Therefore, the need to create a GPUX Dialect arises. The GPUX dialect is an extension of the upstream GPU dialect with following ops - CreateStreamOp, DestroyStreamOp, LaunchFuncOp, AllocOp, DeallocOp & WaitOp . The CreateStreamOp op allows for explicit Stream creation. The stream is a temporary stream and is internally creating a SYCL/L0 queue, with default context and device.
```
silee2, line 12: Although you have already mentioned new ops in the summary above, pointing out which ops are new and which ops are an extension of an existing GPU dialect op would help readers get a better picture.

silee2 (approved):
Looks good to me.

drprajap (approved):

diptorupd:
@nbpatel Thank you for your response.

> For this dialect, I would suggest the GPU dialect just map to underneath Intel GPU, when the higher level compiler choose our GPU pass to lower the GPU dialect, it is to lower to intel GPU. When user dispatch a kernel, it implicitly trigger the compilation targeting “intel GPU” (currently two dispatch/compilation combined, but really should be split). There is no need to accept a “device kind” for the compiler to know the target device type. The current implementation just assumes that it is intel GPU and the device is launched to the default device.

You are missing a fundamental point by making the assumption that the kernel will always be launched on an Intel GPU. As I stated in my comment, it is a behavior that is outside the control of GPUX. At least the way GPUX is being defined. How will you prevent a user from changing the default device to CUDA (refer my example about DPCPP SYCL_DEVICE_FILTER envvar) and causing a run time crash because you are sending a SPIRV kernel to the CUDA driver?

The functionality as stated in https://github.com/intel/mlir-extensions/blob/nishant_rfc/docs/rfcs/GPUXDialect.md#example-usage-in-case-where-users-do-not-provide-the-stream is too error-prone IMO and should not be included at all. 

`GPUToGPUX` pass should be explicit in all ops that it converts. A user has to be absolutely explicit with requirements for the pass to generate any thing. GPUX making any assumptions and allocating on a temporary queue on a "default" device is the wrong thing to do. GPUX should only convert GPU ops when the user has explicitly specified a queue and additionally add a runtime check if the queue is a SPIR-V based Intel GPU device. That is, if you really want to restrict support to Intel GPUs. 

> I changed that. The earlier example was incorrect.
> If users want to use their own stream, we expect the users to write their own custom pass and change the IR to use the stream with our GPUX dialect.
> Users can write a pass to change the function signature to take an additional argument for stream and use the GPUX dialect ops to launch the kernel with their own provided stream. We don't intend on providing that pass as the use case can be varied and depend on users need.

What is the reason for not having the region op in GPUX? Why do we want every upstream dialect to implement the feature themselves? 

If GPUX provides the region op, all an upstream dialect will need to do is generate GPU ops and then wrap those ops with the GPUX region op passing in the stream (kind of like an OMP pragma). `GPUToGPUX` then cleanly translates the GPU ops inside these regions to GPUX ops.

diptorupd:
>> We definitely need memcpy and memset in GPUX.

   > Why? We are only interested in extending ops from upstream GPU dialect which require stream. Memset and memcpy work on memrefs and dont require stream.

My assumption is that these are a wrappers for [`cudamemcpy`](https://developer.download.nvidia.com/compute/DevZone/docs/html/C/doc/html/group__CUDART__MEMORY_g48efa06b81cc031b2aa6fdc2e9930741.html) and [`cudaMemset`](https://developer.download.nvidia.com/compute/DevZone/docs/html/C/doc/html/group__CUDART__MEMORY_ge07c97b96efd09abaeb3ca3b5f8da4ee.html#ge07c97b96efd09abaeb3ca3b5f8da4ee). 

The corresponding SYCL ops all belong to queue class and thus will need a queue argument to be called.

docs/rfcs/GPUXDialect.md
```diff
@@ -0,0 +1,188 @@
...
+
+Create the Context.
+
+Syntax:
+
+```
+operation ::= gpux.create_context $context : (!gpux.ContextType) -> !gpux.ContextType
+```
+
+The gpux.create_context op takes a pointer (opaque) to the underlying context and will return a sycl/l0 context handle.
```
diptorupd, line 43: Nit
```suggestion
The gpux.create_context op takes a pointer (opaque) to an underlying context and will return a sycl/l0 context handle.
```

docs/rfcs/GPUXDialect.md
```diff
@@ -0,0 +1,188 @@
...
+
+Create the GPU Device.
+
+Syntax:
+
+```
+operation ::= gpux.create_device $device : (!gpux.DeviceType) -> !gpux.DeviceType
+```
+
+The gpux.create_device op takes a pointer (opaque) to the underlying device and will return a sycl/l0 device handle.
```
diptorupd, line 30:
```suggestion
The gpux.create_device op takes a pointer (opaque) to an underlying device and will return a sycl/l0 device handle.
```

docs/rfcs/GPUXDialect.md
```diff
@@ -0,0 +1,188 @@
...
+
+Destroy the Device.
+
+Syntax:
+
+```
+operation ::= gpux.destroy_device $device : (!gpux.DeviceType) -> ()
+```
+
+The gpux.destroy_device operation will destroy the above created device.
```
diptorupd, line 69:
```suggestion
The gpux.destroy_device operation will deallocate the passed in device pointer.
```

docs/rfcs/GPUXDialect.md
```diff
@@ -0,0 +1,188 @@
...
+
+Destroy the Context.
+
+Syntax:
+
+```
+operation ::= gpux.destroy_context $context : (!gpux.ContextType) -> ()
+```
+
+The gpux.destroy_context operation will destroy the above created context.
```
diptorupd, line 82:
```suggestion
The gpux.destroy_context operation will deallocate the passed in context pointer.
```

docs/rfcs/GPUXDialect.md
```diff
@@ -0,0 +1,188 @@
...
+
+Destroy the GPU Stream.
+
+Syntax:
+
+```
+operation ::= gpux.destroy_stream $gpux_stream : (!gpux.StreamType) -> ()
+```
+
+The gpux.destroy_stream() operation will destroy the above created stream. If the user wants to use their stream, we expect the user to manage the lifetime of the stream.
```
diptorupd, line 95: Same change as `destroy_device` and `destroy_context` ops.

docs/rfcs/GPUXDialect.md
```diff
@@ -0,0 +1,188 @@
...
+GPU memory allocation operation.
+
+Syntax:
+
+```
+operation ::= gpux.alloc custom<AsyncDependencies>(type($asyncToken), $asyncDependencies)
+              ($gpux_stream, $dynamicSizes) ([ $symbolOperands^ ])? attr-dict : type($memref)
+```
+
+This op is an extension of upstream gpu.alloc op with one added argument for stream.
```
diptorupd, line 109:
```suggestion
This op is an extension of upstream gpu.alloc op with one added argument for stream using which the memory is allocated on a specific device.
```

docs/rfcs/GPUXDialect.md
```diff
@@ -0,0 +1,188 @@
...
+GPU memory deallocation operation.
+
+Syntax:
+
+```
+operation ::= gpux.dealloc custom<AsyncDependencies>(type($asyncToken), $asyncDependencies)
+              ($gpux_stream, $memref) attr-dict : type($memref)
+```
+
+This op is an extension of upstream gpu.dealloc op with one added argument for stream.
```
diptorupd, line 123: Same feedback as alloc.

docs/rfcs/GPUXDialect.md
```diff
@@ -0,0 +1,188 @@
...
+operation ::= gpux.launch_func custom<AsyncDependencies>(type($asyncToken), $asyncDependencies)
+              $gpux_stream,
+              $kernel,
+              blocks in  ($gridSizeX, $gridSizeY, $gridSizeZ)
+              threads in ($blockSizeX, $blockSizeY, $blockSizeZ)
+              (dynamic_shared_memory_size $dynamicSharedMemorySize^)?
+              custom<LaunchFuncOperands>($operands, type($operands)) attr-dict
+```
+
+This op is an extension of upstream gpu.launch_func op with one added argument for stream.
```
diptorupd, line 142:
```suggestion
This op is an extension of upstream gpu.launch_func op with one added argument for stream on which the kernel is enqueued or launched.
```

docs/rfcs/GPUXDialect.md
```diff
@@ -0,0 +1,188 @@
...
+GPU memcpy operation
+
+Syntax:
+
+```
+operation ::= gpu.memcpy custom<AsyncDependencies>(type($asyncToken), $asyncDependencies)
+              $gpux_stream, $dst, $src : type($dst), type($src) attr-dict
+```
+
+The gpux.memcpy operation copies the content of one memref to another. This op is an extension of upstream gpu.memcpy op with one added argument for stream.
```
diptorupd, line 169: Again, some justification is needed for why the stream arg is needed.

diptorupd (approved):
Thanks for incorporating all the feedback and working on improving the RFC through the review process.

I have left some final comments on few awkward phrases that should be fixed either now or before posting an upstream RFC.

Hardcode84:
This context and device types feels completely redundant to me and they too much tied to SYCL concepts. We have zero chances to upstream it in such form IMO, we cannot even reuse it for L0 lowering as L0 context means different thing than SYCL.
L0 context requires driver pointer and we cannot create it from thin air.

What we should have instead is just 3 versions of `create_stream`: version which takes opaque pointer, version with device name/filter-string, and one which takes nothing. And even from them we can realistically only upstream version without parameters.

nbpatel:
Context takes an opaque pointer so :
for SYCL runtime we cast that opaque pointer to device, since sycl can do something like sycl::context context($device)
for level zero we can cast that opaque pointer to ze_driver_handle_t and use that to create context for level zero

Hardcode84:
> Context takes an opaque pointer so :
for SYCL runtime we cast that opaque pointer to device, since sycl can do something like sycl::context context($device)
for level zero we can cast that opaque pointer to ze_driver_handle_t and use that to create context for level zero

This still doesn't answer why we need separate context concept. It is too much tied to SYCL concepts and non-upstreamable.
And if you have `create_stream` taking opaque pointer you can hide all those context manipulations behind it. 

Also, ping @diptorupd

drprajap (commented):
LGTM, minor description change might be needed..

Hardcode84:
LGTM to get things going. We still can revise it when we decide to propose RFC upstream.

Jianhui-Li:
> LGTM to get things going. We still can revise it when we decide to propose RFC upstream.

Agree. Certainly, I expect the RFC will evolve.

PR #329, opened 2022-09-18
Update llvm sha

Hardcode84:
Changes:
* C++ version switched to C++17
* Default accessors format switched to camelCase
* `llvm::array_lengthof` was removed
* `LLVM::Extract/InsertValueOp` builder now accepts plain integers
* spirv ops naming changes


Hardcode84:
@fschlimb @diptorupd @nbpatel FYI

PR #334, opened 2022-09-26
IMEX_BINARY_DIR differs depending on whether IMEX is built standalone vs

silee2:
as an LLVM_EXTERNAL_PROJECT.
pass pipeline files do not need to be copied to binary directory.

Current llvm external project build for imex has an issue with copying some files.
This PR fixes the root cause of that issue.

LLVM external project build places libraries and executables under CMAKE_BINARY_DIR and not is CMAKE_CURRENT_BINARY_DIR.
Note that MLIR tablegen generated headers are always put relative to CMAKE_CURRENT_BINARY_DIR
This PR addresses issue coming from that difference.

Supplementary files in test directory can be accessed in lit run script by using "%p" substitution variable. The variable gives path to current test directory. And can be used like the following.
// RUN: ... %p/<somefilename> ...

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -50,9 +50,6 @@ else()
     list(APPEND CMAKE_MODULE_PATH "${LLVM_CMAKE_DIR}")
 endif()
 
-set(IMEX_SOURCE_DIR ${CMAKE_CURRENT_SOURCE_DIR})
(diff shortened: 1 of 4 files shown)


silee2:
@mshahneo Can not find you in the pop up list.

drprajap (approved):

silee2:
Merging since the PR got two approvals.

PR #335, opened 2022-09-27
Rename includes `mlir-extension` -> `imex`

Hardcode84:
No functional changes


Hardcode84:
@diptorupd @fschlimb FYI

PR #337, opened 2022-09-29
Port GPUToSPIRV Pass

drprajap:
This pass extends upstream GPUToSPIRV pass by adding more conversion patterns to include
    SCF, arith, math ops within gpu kernel module.


drprajap:
@Jianhui-Li and @mshahneo , cannot find you both in reviewers popup list.

Jianhui-Li:
Looks good to me.

docs/Conversion/GPUToSPIRV.md
```diff
@@ -0,0 +1,152 @@
+# GPUXToSPIRVPass
```
fschlimb, line 1: The recommended way of providing docs is in .td files. There is a whole machinery to extract and assemble/group docs from various places to generate a consistent doc structure. Individual doc files counteract this approach.

docs/Conversion/GPUToSPIRV.md
```diff
@@ -0,0 +1,152 @@
...
+        spv.mlir.merge
+      }
+      spv.Return
+    }
+  }
+
+
+## Reason for this Custom Pass:
+Upstream pass includes limited conversion patterns and it does not add patterns conversion from SCF, arith and math ops.
+So GPUXToSPIRV pass extends updatream pass conversion.
```
fschlimb, line 152: typo

include/imex/Conversion/GPUToSPIRV/GPUToSPIRVPass.h
```diff
@@ -0,0 +1,37 @@
+//===- GPUToSPIRVPass.h - GPUToSPIRV conversion ------------------------------*-
+// C++ -*-===//
```
fschlimb, line 2: should be one line

docs/Conversion/GPUToSPIRV.md
drprajap: I have added description in .td file as well, but because I added some examples for this pass extension and that is a bit lengthy, that would have clobbered .td file with unnecessary details, so thought of adding them in separate .md file.

lib/Conversion/GPUToSPIRV/GPUToSPIRVPass.cpp
```diff
@@ -0,0 +1,122 @@
...
+/// pass only converts gpu.func ops inside gpu.module op.
+///
+//===----------------------------------------------------------------------===//
+#include "imex/Conversion/GPUToSPIRV/GPUToSPIRVPass.h"
+
+#include "../PassDetail.h"
+#include "mlir/Conversion/MemRefToSPIRV/MemRefToSPIRV.h"
+#include "mlir/Dialect/GPU/IR/GPUDialect.h"
+
+#include <mlir/Conversion/ArithmeticToSPIRV/ArithmeticToSPIRV.h>
```
Hardcode84, line 22: I think some of those includes can be removed.

docs/Conversion/GPUToSPIRV.md
fschlimb: Look at upstream passes, most of them (not all, though) have all the docs in the .td file, no matter how long they are (like https://github.com/llvm/llvm-project/blob/129420df5122a27560a17f6a472c8c750d3bbd7b/mlir/include/mlir/Dialect/Bufferization/Transforms/Passes.td).

Spreading docs across multiple places makes maintenance even worse. Less files, less policies, less trouble.

What's the point of having an extra file?

silee2 (approved):
LGTM

nbpatel (approved):

lib/Conversion/GPUToSPIRV/GPUToSPIRVPass.cpp
drprajap: Done

docs/Conversion/GPUToSPIRV.md
drprajap: I didn't see one of these .td file which included example input and output so assumed td file documentation does not include that level of detail. Thanks for sharing above, I have removed this extra doc file and added examples in .td file.

include/imex/Conversion/GPUToSPIRV/GPUToSPIRVPass.h
```diff
@@ -0,0 +1,36 @@
...
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+///
+/// \file
+/// This file defines the GPUToSPIRV conversion,
+///
+//===----------------------------------------------------------------------===//
+
+#ifndef IMEX_GPUTOSPIRV_PASS_H_
```
Hardcode84, line 15: Side question: are we going with include guards in refactor or with `#pragma once`?

Hardcode84 (approved):

include/imex/Conversion/GPUToSPIRV/GPUToSPIRVPass.h
drprajap: so far I have seen upstream is following include guards rather than #pragma once, and refactor is also following the same I believe, @fschlimb/others can clarify if there are other plans?

fschlimb: Yes, I also see upstream using #ifndef guards. They are more flexible, more reliable and standard C++. I guess both approaches would be acceptable in most cases, though. I don't know if there was a explicit decision made in upstream but I suggest we follow what happens there (only clang seems to sometimes use the pragma).

Issue #316, opened 2022-09-09
[ptensor] add missing elementwise binary operations


fschlimb:
The following elementwise binary operations need an implementation in PTensorToLinalg(.cpp):
* [ ] ptensor::ATAN2
* [ ] ptensor::LOGADDEXP
* [ ] ptensor::LSHIFT
* [ ] ptensor::MATMUL
* [ ] ptensor::TRUE_DIVIDE
* [ ] ptensor::BITWISE_AND
* [ ] ptensor::BITWISE_LEFT_SHIFT
* [ ] ptensor::BITWISE_OR
* [ ] ptensor::BITWISE_RIGHT_SHIFT
* [ ] ptensor::BITWISE_XOR
* [ ] ptensor::EQUAL
* [ ] ptensor::GREATER
* [ ] ptensor::GREATER_EQUAL
* [ ] ptensor::LESS
* [ ] ptensor::LESS_EQUAL
* [ ] ptensor::LOGICAL_AND
* [ ] ptensor::LOGICAL_OR
* [ ] ptensor::LOGICAL_XOR
* [ ] ptensor::NOT_EQUAL

See also [Operation Details->Elementwise Operations in the RFC](https://github.com/intel/mlir-extensions/tree/refactor/docs/rfcs/20220804-ptensor#operation-details).

Reference implementations can be found in the TOSA dialect and/or on main (numba_dpcomp/numba_dpcomp/mlir/numpy/funcs.py [Python]).

It is ok to let initial implementations operate on default PTensorTypes only, e.g. ignore device and distribution attributes of input tensors.

PR #344, opened 2022-10-03
Added all possible elementwise binary ops for #316.

chudur-budur:
    - A draft PR for #316. Added all possible element-wise binary ops.
    - The same test script should work with this.
    - Yes
   - There are some new warnings, but they are not mitigated yet.
    - Yes
    - Yes
    
@diptorupd @fschlimb


lib/Conversion/PTensorToLinalg/PTensorToLinalg.cpp
```diff
@@ -293,19 +295,63 @@ static BodyType buildTrivial(::mlir::Type typ) {
   };
 }
 
+// Builder for TOSA body.
+// The builder function takes an extra agrument ::mlir::Type
+// Many TOSA functions take two arguments and do the
+// same operation on both arguments.
+// TODO:
+//      1. Find a way to merge this function with trivialBuilder().
```
fschlimb, line 303: Why would we need buildTrivial if we use TOSA?

lib/Conversion/PTensorToLinalg/PTensorToLinalg.cpp
```diff
@@ -293,19 +295,63 @@ static BodyType buildTrivial(::mlir::Type typ) {
   };
 }
 
+// Builder for TOSA body.
+// The builder function takes an extra agrument ::mlir::Type
+// Many TOSA functions take two arguments and do the
+// same operation on both arguments.
```
fschlimb, line 301: Is this the functionality we need? Why is TOSA any better than math and arith?

include/imex/Dialect/PTensor/IR/PTensorOps.h
```diff
@@ -53,6 +53,7 @@ enum EWBinOpId : int {
   NOT_EQUAL,
   OR,
   POWER,
+  RSHIFT,
```
fschlimb, line 56: I guess LSHIFT and RSHIFT are duplicates of BITWISE_LEFT_SHIFT and BITWISE_RIGHT_SHIFT.

fschlimb:
Please also note that the RFC makes the following distinction:
* `elementwise_binary_op{$ebop}(lhs, rhs) : (ptensor.ptensor, ptensor.ptensor) -> ptensor`
    * `$ebop = ['add', 'sub', 'bitwise_left_shift', ...]`
* `elementwise_binary_test{$ebop}(lhs, rhs) : (ptensor.ptensor, ptensor.ptensor) -> ptensor`
  * `$ebop = ['greater', 'less', 'equal', 'logical_and', ...]`

lib/Conversion/PTensorToLinalg/PTensorToLinalg.cpp
chudur-budur: That's my thought too, should we just move to TOSA completely?

lib/Conversion/PTensorToLinalg/PTensorToLinalg.cpp
chudur-budur: @fschlimb I found those logical ops only in TOSA, is there any other alternative?

include/imex/Dialect/PTensor/IR/PTensorOps.h
chudur-budur: @fschlimb Okay, we can just duplicate them.

chudur-budur:
> Please also note that the RFC makes the following distinction:
> 
> * `elementwise_binary_op{$ebop}(lhs, rhs) : (ptensor.ptensor, ptensor.ptensor) -> ptensor`
>   
>   * `$ebop = ['add', 'sub', 'bitwise_left_shift', ...]`
> * `elementwise_binary_test{$ebop}(lhs, rhs) : (ptensor.ptensor, ptensor.ptensor) -> ptensor`
>   
>   * `$ebop = ['greater', 'less', 'equal', 'logical_and', ...]`

Hi, I couldn't find this RFC, where is it?

fschlimb:
> Hi, I couldn't find this RFC, where is it?

docs/rfcs/20220804-ptensor/README.md

include/imex/Dialect/PTensor/IR/PTensorOps.h
fschlimb: Well, I actually thought we should not. Any reason for doing so?

lib/Conversion/PTensorToLinalg/PTensorToLinalg.cpp
fschlimb: I don't think so. An alternative is of course to write them manually.

lib/Conversion/PTensorToLinalg/PTensorToLinalg.cpp
fschlimb: When I investigated I realized that TOSA does not have unsigned types. I could not figure out if that's a gap or a design decision. If we can assume that TOSA can support unsigned ints then I the only reason for not using TOSA might be compile-time. But I tend to think that's probably acceptable at this point.

include/imex/Dialect/PTensor/IR/PTensorOps.h
chudur-budur: Well, actually there was `LSHIFT` but no 'RSHIFT' so I added it.

chudur-budur:
> > Hi, I couldn't find this RFC, where is it?
> 
> docs/rfcs/20220804-ptensor/README.md

What is the difference between `elementwise_binary_op` and `elementwise_binary_test`? Not clear to me.

fschlimb:
> What is the difference between `elementwise_binary_op` and `elementwise_binary_test`? Not clear to me.

The tests return a boolean tensor always. The others return numbers. If we do not need excessive conditionals in the implementation it is probably ok to combine.

include/imex/Dialect/PTensor/IR/PTensorOps.h
fschlimb: I prefer BITWISE_*_SHIFT.

chudur-budur:
@fschlimb I think it's ready for merge.

fschlimb:
Thanks @chudur-budur 
I guess we need tests. I suggest defering this until we resolved #380

PR #347, opened 2022-10-04
Make gpu barriers follow uniform control flow

Hardcode84:
gpu barriers require uniform control flow, but in our kernel api we always insert top level `IfOp` to avoid out of bounds access if global and local sizes are not evenly divisible.

Add new pass which splits such `IfOp` in a way to put barrier in uniform flow, see tests for specific transformation details.

Also:
* This pass exposed some additional issues
* `gpu_runtime::StorageClassAttr` is not handled well by upstream memref type converter, use `mlir::gpu::GPUDialect::getPrivateAddressSpace()` for now (need to fix upstream)
* Add new `util.sign_cast` canonicalization as it used as memory space cast op now (THIS IS A HACK, need dedicated op for this)


Hardcode84:
@diptorupd @nbpatel @silee2 FYI

PR #348, opened 2022-10-05
Fix GPU to spirv subgroup builtins lowering

Hardcode84:
Upstream GPU to spirv lowering controls builtins type lowering via single variable `use64bitIndex`, but IGC expects get_global/local/id/size to be i64 (size_t) but subgroup builtins are still expected to be i32 (uint).

For now, set `use64bitIndex=true` but add custom lowering for subgroup builtins to lower them to i32.

### DIFF
--- a/mlir/lib/Conversion/gpu_to_gpu_runtime.cpp
+++ b/mlir/lib/Conversion/gpu_to_gpu_runtime.cpp
@@ -1115,6 +1115,35 @@ public:
   }
 };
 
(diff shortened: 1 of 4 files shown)


Hardcode84:
@drprajap @nbpatel @silee2 FYI

drprajap:
thanks for letting us know Ivan, I will cherry-pick your change to refactor once it is merged.

PR #350, opened 2022-10-07
Add select set of Argonne Jax test lowered to linalg dialect.

silee2:
This is the initial set of mlir test cases from Argonne Jax
Lit test is disabled for the new directory by updating lit.cfg.py
Tests are raw mlir files without any RUN command or CHECK directives.
Test enabling will be done incrementally in the future.
Input and golden output needs to be added to test cases to enable end to end execution.
Since this is the initial set, more test may be added in the future.
Also added resnet-50 and mobilenet-v3 models in linalg dialect.


mshahneo (commented):
Looks good to me.
One quick question, @silee2 : Do all these tests run? The reason I am asking is that, will they be part of CI?

silee2:
> Looks good to me. One quick question, @silee2 : Do all these tests run? The reason I am asking is that, will they be part of CI?

They won't be part of CI initially. The tests are meant from end to end flow which is still under development for refactor branch. lit.cfg.py excludes these new directory from testing. They will be enabled over time.

Dewei-Wang-sh (approved):
lgtm

silee2:
> Can we remove the module__* from the file names and just have the functionality to make it more readable?

Removed.

PR #382, opened 2022-10-17
Update llvm sha

Hardcode84:
Need to pull following fixes from upstream:
* https://reviews.llvm.org/D135323 `gpu.subgroup_reduce` op
* https://reviews.llvm.org/D133533 `host_shared` flag to `gpu.alloc`
* https://reviews.llvm.org/D135907 properly handle case when `TargetEnv` attached to  `gpu.module` instead of top-level module in gpu-to-spirv lowering
* https://reviews.llvm.org/D135725 fix crash in `linalg.generic` canonicalized


Hardcode84:
@nbpatel @silee2 @fschlimb @diptorupd FYI

PR #384, opened 2022-10-18
Properly check L0 device capabilities before running gpu kernels

Hardcode84:
This is needed to avoid crashes when running on devices without f64 support.

* Set `TargetEnvAttr` per each `gpu.module` separately instead of top-level module as different gpu modules can potentially have different capabilities.
* Add callback to `createSetSPIRVCapabilitiesPass` so user can provide custom capabilities instead of hardcoding them.
* Add `dpcompGetDeviceCapabilities` to gpu runtime to get device caps
* Connect passes to gpu runtime using python bridge.


Hardcode84:
@nbpatel @silee2 @drprajap @diptorupd FYI

PR #389, opened 2022-10-19
Adding ptensor.extract_slice

fschlimb:
New Op extract_slice as a shallow wrapper around tensor.extract_slice plus lowering and tests.
Not handled in Dist Dialect yet (watch out for extra PR).
Similar to rest of PTensor no device handling yet.


Hardcode84:
Nit: slide-> slice in description

lib/Conversion/PTensorToLinalg/PTensorToLinalg.cpp
```diff
@@ -156,6 +153,40 @@ struct ExtractRTensorLowering
...
+    auto srcRtTyp = srcPtTyp.getRtensor();
+    auto retPtTyp = op.getType().dyn_cast<::imex::ptensor::PTensorType>();
+    if (!retPtTyp)
+      return ::mlir::failure();
+    auto retRtTyp = retPtTyp.getRtensor();
+    // extract RankedTensor
+    auto source = rewriter.create<::imex::ptensor::ExtractRTensorOp>(
+        loc, srcRtTyp, op.getSource());
+    // and simply replace with tensor::extract_slice
+    rewriter.replaceOpWithNewOp<::mlir::tensor::ExtractSliceOp>(
```
Hardcode84, line 181: Just FYI: we cannot blindly convert all ptensor/ntensor slices to `tensor.extract_slice` as they are reference types and `tensor` is value type and is immutable and some tensor/linalg transformations rely on that. We need some sort of alias analysis to make sure we have no writes (e.g. setitem) to `tensor.extract_slice` result, but this doesn't implemented in ntensor either yet.

But this PR is still ok to as as starting point.

fschlimb: I guess you mean writing to the tensor with operations which are not in the tensor dialect. Is that what you mean?

Hardcode84: Consider following example:
```
array = numpy.zeros(...)
view = array[1:2]
view[0] = 1
print(array[1])
```

Naive conversion to linalg will produce smth like this
```
%0 = linalg.fill ...
%1 = tensor.extract_slice %0, ...
setitem(%1, ...)
%3 = getitem %0
```
And as tenors assumed immutable, it can be converted to
```
%0 = linalg.fill ...
%1 = linalg.fill ...
setitem(%1, ...)
%3 = getitem %0
```

fschlimb: Ok. 
I remember that the docs for tensor.insert_slice gave the impression that it provides inplace updates. It now clearly states that it returns a copy, so that is not a way out either.
How would we guard writes from such undesired optimizations? Is there something like an annotation which we can add?

Hardcode84: For ntensor I'm now also always convert them, but in the future I'm planning to generate `tensor.extract_slice` only if we can prove it is safe (no writes to result), otherwise it will stay as `ntensor.subview` and will be converted to `memref.subview` later. As I said, it require alias analysis which I need to think more of.

include/imex/Dialect/Dist/IR/DistOps.td
```diff
@@ -30,12 +30,41 @@ def Dist_Dialect : Dialect {
...
+}
+
+def Dist_Info : Dist_Type<"DistInfo", "info">
+{
+  let summary = "A type for attaching distributed info information to a PTensor";
+  let description = [{
+    A distributed PTensor needs info information like offset and shape of local partition.
+    The DistInfo type is used to define operations to carry and extract such info data.
+  }];
+  let parameters = (ins "int32_t":$rank);
```
chudur-budur, line 59: Is `rank` similar to the ranks of process in a group like in MPI?

include/imex/Dialect/Dist/IR/DistOps.td
```diff
@@ -46,48 +75,149 @@ def Dist_Dialect : Dialect {
...
-// Add function prototypes used for calling into distributed runtime
 def RuntimePrototypesOp : Dist_Op<"runtime_prototypes"> {
+    let summary = "Add function prototypes used for calling into distributed runtime";
 }
 
-// Register a ptensor of given shape with a (potentially distributed) runtime.
-// Returns an id to uniquely identify the tensor instance in future interactino with the runtime.
-// The runtime does not own or manage any PTensor memory. When needed by an operation,
-// (local) data needs to be provided.
-def RegisterPTensorOp : Dist_Op<"register_ptensor", []> {
```
chudur-budur, line 57: Why is `RegisterPTensorOP` removed?

include/imex/Dialect/Dist/IR/DistOps.td
```diff
@@ -46,48 +75,149 @@ def Dist_Dialect : Dialect {
...
-    // Global shape needed for initial registration. Views are handled by a separate op.
-    let arguments = (ins AnyType: $shape);
+def NProcsOp : Dist_Op<"nprocs", [Pure]> {
+    let summary = "Number of processes for given team";
+    let arguments = (ins AnyType:$team);
```
chudur-budur, line 84: Isn't `$team` supposed to be an integer number? Why is it `AnyType`?

include/imex/Dialect/Dist/IR/DistOps.td
```diff
@@ -46,48 +75,149 @@ def Dist_Dialect : Dialect {
...
+    let builders = [
+        OpBuilder<(ins "::mlir::Value":$team), [{
+            build($_builder, $_state, $_builder.getI64Type(), team);
+        }]>,
+    ];
 }
 
-// Get the offsets (one for each dimension) of the local partition of a distributed PTensor in number of elements.
-// Partitionings can be N-dimensional but must cut only the first N dimensions.
-def LocalOffsetsOp : Dist_Op<"local_offsets", []> {
```
chudur-budur, line 67: Why this has been removed?

include/imex/Dialect/Dist/IR/DistOps.td
```diff
@@ -46,48 +75,149 @@ def Dist_Dialect : Dialect {
...
+    let summary = "Initialize distributed meta information about a new PTensor.";
+    let description = [{
+        The distributed meta information about a new PTensor provides
+          - the rank of the tensor
+          - the process-local shape
+          - the process-local offsets
+          - the distributed team
+    }];
+    // Global shape needed for initial registration. Views are handled by a separate op.
+    let arguments = (ins I64Attr:$rank, AnyType:$shape, AnyType:$team);
```
chudur-budur, line 114: Why are `$team` and `$shape` `AnyType`?

include/imex/Dialect/Dist/IR/DistOps.td
```diff
@@ -46,48 +75,149 @@ def Dist_Dialect : Dialect {
...
+      }]>,
+    ];
 }
 
-// Get the shape (one size for each dimension) of the local partition of a distributed PTensor in number of elements.
-// Partitionings can be N-dimensional but must cut only the first N dimensions.
-def LocalShapeOp : Dist_Op<"local_shape", []> {
-    // Id of tensor as returned by RegisterPTensorOp
-    let arguments = (ins I64Attr: $rank, I64: $ptensor);
+def GetInfoOp : Dist_Op<"get_info", [Pure]> {
```
chudur-budur, line 143: What is the difference between `GetInfoOp` and `DistInfoOp`?

lib/Dialect/PTensor/Transforms/PTensorDist.cpp
```diff
@@ -48,32 +54,29 @@ namespace {
...
-// create ::imex::dist::RegisterPTensorOp
-inline ::mlir::Value createInitWithRT(::mlir::Location &loc,
-                                      ::mlir::OpBuilder &builder, uint64_t rank,
-                                      ::mlir::Value gshape) {
-  return builder.create<::imex::dist::RegisterPTensorOp>(
-      loc, builder.getI64Type(), gshape);
+// create ::imex::dist::DistInfoOp
+inline ::mlir::Value createDistInfo(::mlir::Location &loc,
+                                    ::mlir::OpBuilder &builder, uint64_t rank,
+                                    ::mlir::Value gshape, ::mlir::Value team) {
```
chudur-budur, line 60: I am assuming `gshape` means shape of a group, but it's not very clear what do we exactly mean by that.

chudur-budur (commented):
I need to read this code more. Most of the stuffs are not within my understanding yet.

include/imex/Dialect/Dist/IR/DistOps.h
```diff
@@ -26,7 +26,7 @@ namespace ptensor {
 class PTensorType;
 } // namespace ptensor
 namespace dist {
-enum INFO : int { LSHAPE, LOFFSETS, TEAM, INFO_LAST };
+enum INFO : int { GSHAPE, LSHAPE, LOFFSETS, TEAM, INFO_LAST };
```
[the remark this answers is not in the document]
fschlimb: *g*lobal *s*hape

include/imex/Dialect/Dist/IR/DistOps.td
fschlimb: In MLIR `rank` refers to the number of dimensions of a tensor. This is what is meant here.

You are right, we have a clash of terminology here. In later version I use `prank` for the process-rank.

include/imex/Dialect/Dist/IR/DistOps.td
fschlimb: In my pass implementation it is an int, some other implementation might use something else (like an ip-address).

include/imex/Dialect/Dist/IR/DistOps.td
fschlimb: Shapes can be represented in various ways, like as a `shape.shape` or `tensor` (with different element types) or memref.

include/imex/Dialect/Dist/IR/DistOps.td
fschlimb: `DistInfoOP` is the initialization/creation of an `DistInfo` from tensor + shape + team. `GetInfoOp` returns the `DistInfo` from a given `DistTensor`.

include/imex/Dialect/Dist/IR/DistOps.td
fschlimb: The new dialect no longer relies on a runtime to compute offsets, local shapes etc. PAsses generate code to compute in inline. This op got replace by `InitDistTensorOp`.

include/imex/Dialect/Dist/IR/DistOps.td
fschlimb: Local offsests are now provided by the `DistInfoOp` and can be retrieved through `ExtractFromInfoOp`.

lib/Dialect/PTensor/Transforms/PTensorDist.cpp
fschlimb: see above, it's the global shape of the tensor (in contrast to the process-local shape).

fschlimb:
This has become a rather large PR deviating from the original intent.

The bulk of the changes address the following
- the ptensor dialect no longer deals with any distribution semantics
- an extra pass adds the necessary ops needed for distributed operation and then calls the "normal" ptensor op, potentially followed by some distributed after-care
- Compute-follows data is enabled by introducing a new DistTensorType
  - to create a DIstTensor (like by an arange) the operation annotates the PTensor with distributed meta information (such as global shape and local offsets)
  - Later, the dist-to-standard pass extracts the necessary meta information from that operation and lowers to the appropriate ops
- on function boundaries DistTensors get expanded to multiple args (gshape, ltensor, loffsets, team)

This works with my jited, distributed tensor library.

Not sure if this is how @Hardcode84 thought this could work, but conceptually we could do the same for devices and so separate tensor-semantics from how they are mapped to HW/backends.

include/imex/internal/PassUtils.h
```diff
@@ -18,11 +18,17 @@
 
 namespace imex {
 
+/// @return get ::mlir::IntegerAttr with given Value and bitwidth W
+template <int W = 64, typename T = int64_t>
```
Hardcode84, line 22: nit: `getIntegerAttr` takes `int64_t` anyway so this `typename T` useless

include/imex/Dialect/PTensor/IR/PTensorOps.td
```diff
@@ -126,10 +122,60 @@ def ExtractRTensorOp : PTensor_Op<"extract_rtensor", []> {
...
+        auto sourcePTensorType = source.getType().cast<PTensorType>();
+        auto resultPTensorType = source.getType().cast<PTensorType>();
+        if(sourcePTensorType && resultPTensorType) {
+            ::mlir::tensor::ExtractSliceOp::build(odsBuilder, odsState, resultPTensorType, source, std::forward<Ts>(args)...);
+        } else {
+            // FIXME
+            assert("expected PTensor(s) in ptensor.extract_slice" == nullptr);
+        }
+    }
+  }];
```
Hardcode84, line 176: It is missing `infer(RankReduced)ReturnType` functions present in upstream tensor/memref counterparts

lib/Conversion/DistToStandard/DistToStandard.cpp
```diff
@@ -31,23 +35,35 @@
...
+/// After full conversion the cast is expected
+/// to have no use. FIXME Is there a better/cleaner way to do this?
+::mlir::Value materializeDistTensor(::mlir::OpBuilder &builder,
+                                    ::mlir::Location loc, ::mlir::Value gshape,
+                                    ::mlir::Value ltensor,
+                                    ::mlir::Value loffsets,
+                                    ::mlir::Value team) {
+  // Put named arguments into given Vector in a well defined order,
+  // so that extraction is correct
+  ::mlir::SmallVector<::mlir::Value> vals(::imex::dist::INFO_LAST);
```
Hardcode84, line 49: `std::array` as size is known at compile time.

lib/Conversion/DistToStandard/DistToStandard.cpp
```diff
@@ -102,79 +123,168 @@ struct RuntimePrototypesOpConverter
...
-                  ::mlir::PatternRewriter &rewriter) const override {
-    // get guid and rank and call runtime function
+                  ::imex::dist::LocalShapeOp::Adaptor adaptor,
+                  ::mlir::ConversionPatternRewriter &rewriter) const override {
+    // FIXME: non-even partitions, ndims
     auto loc = op.getLoc();
-    rewriter.replaceOp(
-        op, createCallGetRankedData(loc, rewriter, "_idtr_local_shape",
-                                    op.getPtensor(), op.getRank()));
+    auto converter = *getTypeConverter();
```
Hardcode84, line 254: This will do a full copy of `converter ` which is probably not what you wanted.

Hardcode84:
> Not sure if this is how @Hardcode84 thought this could work, but conceptually we could do the same for devices and so separate tensor-semantics from how they are mapped to HW/backends.

You basically did inverse of what I proposed, `Device<Distributed<Tensor<?xf32>>, "smth">` vs `Tensor<?xf32, Distributed<Device<"smth">>>>` but this can work too, I guess

fschlimb:
> > Not sure if this is how @Hardcode84 thought this could work, but conceptually we could do the same for devices and so separate tensor-semantics from how they are mapped to HW/backends.
> 
> You basically did inverse of what I proposed, `Device<Distributed<Tensor<?xf32>>, "smth">` vs `Tensor<?xf32, Distributed<Device<"smth">>>>` but this can work too, I guess

Ok, sorry, my questions was not super clear. I understand that this is not what you suggest for NTensor. Is the NTensor approach what you had always had in mind? Maybe I misunderstood, but I got the impression that you suggested something different a while back. 

I don't think it is a good idea to make the team a(n indirect) part of the type as explained elsewhere. I guess it needs some discussion to better understand the pros and cons of making it a compile time constant (or at least the #processes).

include/imex/Dialect/Dist/IR/DistOps.td
```diff
@@ -46,48 +68,124 @@ def Dist_Dialect : Dialect {
...
+}
 
-    // result is a 1d memref
+def ExtractFromDistOp : Dist_Op<"extract_from_dist", []> {
+    let summary = "Get one element from the distributed meta information.";
+    let arguments = (ins I32Attr:$what, AnyType:$d_tensor);
     let results = (outs AnyType);
+    let builders = [
+      // auto-deduce return type from from operands
+      OpBuilder<(ins "::imex::dist::INFO":$what, "::mlir::Value":$d_tensor), [{
```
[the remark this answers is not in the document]
fschlimb: Well, when I look at existing .td files I see much more C++ code than what is here. 
Can you point me to something which suggests the recommended practice?

include/imex/Dialect/PTensor/IR/PTensorOps.td
fschlimb: Would it needed/useful for anything other than "internal" use?

include/imex/internal/PassUtils.h
fschlimb: Fixed.

lib/Conversion/DistToStandard/DistToStandard.cpp
fschlimb: done.

lib/Conversion/DistToStandard/DistToStandard.cpp
fschlimb: Fixed.

Hardcode84:
> I don't think it is a good idea to make the team a(n indirect) part of the type as explained elsewhere. I guess it needs some discussion to better understand the pros and cons of making it a compile time constant (or at least the #processes).

I'm not suggesting making team a static part of dist type, I'm not even sure why do we need it in dist type at all, it looks more like implementation detail of specific lowering path.

fschlimb:
> I'm not suggesting making team a static part of dist type, I'm not even sure why do we need it in dist type at all, it looks more like implementation detail of specific lowering path.

The team carries information which cannot be deduced by the compiler; it defines a group of processes which operate on the tensor. Not always all processes will operate on all dist-tensors.

the disttensor does not contain any information other than that the tensor is distributed. The team is provided by generator ops and/or by block arguments only.

If not the team, what else should be in the env-descriptor then?

Hardcode84:
> the disttensor does not contain any information other than that the tensor is distributed. The team is provided by generator ops and/or by block arguments only.

Team is dynamic value and materialized during conversion from dist to low-level dialects if I understand your code correctly, we don't need any additional static information in dist type for this, only the fact that it is distributed

fschlimb:
> Team is dynamic value and materialized during conversion from dist to low-level dialects if I understand your code correctly, we don't need any additional static information in dist type for this, only the fact that it is distributed

Exactly, that's why the original PTensor-type had only a boolean as the dist flag.

For the GPU/device handling it seems the situation is similar in the sense that the device-number is a dynamic attribute but - unless we want fat binaries - the device type should be static. So for devices we probably want the env to be a bit richer than a boolean. It is worth a separate discussion whether the complete separation between Tensor API and execution target (as done here) should also be followed for devices (in contrast to embedding the execution target into the base tensor type as done in PTensor and NTensor).

Hardcode84:
> So for devices we probably want the env to be a bit richer than a boolean.

For L0/SCYL device type actually is not enough, we also need to store device capabilities (max glolbal/local size, local mem size, double support, etc) as they affect codegen, but for now I'm just storing device name (filter string) in the attr.

silee2 (approved):

PR #390, opened 2022-10-19
Skip some parfor tests

Hardcode84:
* Skip tests which check x86 asm, as they are useless for GPU anyway
* Skip tests checking for specific Parallel Accelerator exceptions
* Skip test with `Cache = True`


Hardcode84:
@diptorupd FYI

PR #394, opened 2022-10-21
Fix `InsertGPUAllocs` pass to support memrefs with non-identity layout

Hardcode84:
`gpu.alloc` will allocate memref with default layout which then need to be casted to memref original layout.


Hardcode84:
@nbpatel FYI

PR #403, opened 2022-10-22
Improve gpu-to-gpux pass and surrounding infra

Hardcode84:
* Introduce `StreamType` to `gpu_runtime` dialect instead if opaque type.
* Add `device` attr to `create_stream`
* Add custom assembly to gpu_runtime alloc/dealloc  and create/destroy stream.
* make `gpu-to-gpux` pass aware of env regions. Generate separate create  `create_stream` for ops residing in regions with different device.


Hardcode84:
@diptorupd @nbpatel FYI

PR #405, opened 2022-10-22
Add env region support to InsertGPUAllocs pass

Hardcode84:
If device reads or writes are inside gpu region, corresponding `gpu.alloc` and `memref.copy` will also be wrapped into same region.


Hardcode84:
@diptorupd @nbpatel FYI

PR #408, opened 2022-10-23
Pass device name from `create_stream` to gpu runtime

Hardcode84:
This device name is ignored in runtime for now, will fix in separate PR.


Hardcode84:
@diptorupd @nbpatel FYI

PR #409, opened 2022-10-23
Pass device name to gpu runtime

Hardcode84:
As we still using L0 runtime we need to parse filter string manually. The only supported backend is `level_zero` and the only supported devices are `cpu` and `gpu`.


Hardcode84:
@diptorupd @nbpatel FYI

PR #411, opened 2022-10-24
Add memssa-based mem opt pass and tests

Hardcode84:


Hardcode84:
@Dewei-Wang-sh I've exposed memory opt pass and added vector dialect support (only `vector.load/store` for now but adding `transfer_write/read` should be straightforward). I wasn't able to compile your example, but if you get it working I can add it as test as well.

Issue #289, opened 2022-08-11
cmake config needs fixing


fschlimb:
Currently, on refactor, the following things don't work correctly with cmake:
* lib/cmake/imex/IMEXTargets.cmake is not being installed
  * Ideally we'd do this with LLVM's machinery, but that requires more than just `set_property(GLOBAL PROPERTY IMEX_HAS_EXPORTS TRUE)`.
* include headers are nested in a double 'include' dir (e.g. imex/include/include/imex/...)

PR #415, opened 2022-10-26
Enable end to end test with Linalg input and Vulkan backend

silee2:
This PR enables the first end to end test targeting GPU.
It has multiple related sub parts.

1. Extend insert-gpu-allocs, set-spirv-capabilities and set-spirv-abi-attrs to support both OpenCL and Vulkan SPIRV
2. Change how insert-gpu-allocs, set-spirv-capabilities and set-spirv-abi-attrs passes gets registered.
3. Add test support on Windows
4. Extend imex-runner.py - Support comments in pass pipeline file, add options to print IR before and after passes, add support for mlir-vulkan-runner in addition to mlir-cpu-runner
5. Other compile issues uncovered by building on Windows.
6. Add end to end test case.
7. Change some defaults for gpu to spirv conversion pass. Now spirv memory space mapping is enabled and targets OpenCL.
8. OpenCL pipeline is not completely end to end but includes all passes implemented so far. Upcoming passes are also added to pass pipeline but commented out for now.


silee2:
I'll be taking PTO on Thur 10/17, so response would be delayed.

silee2:
CI seems to have some LLVM caching issue after LLVM SHA update. CI throws error that some files are not found.

fschlimb (approved):
Nice!
A few more comments in the code would be helpful.

fschlimb:
@diptorupd some changes to the cmake machinery, maybe this also touches #289

fschlimb:
> 6\. Add end to end test case.

end-to-end would start with PTensor :)

lib/Transforms/SetSPIRVAbiAttribute.cpp
```diff
@@ -16,40 +16,75 @@
...
     auto gpuModule = getOperation();
     auto *context = &getContext();
     auto attrName =
         mlir::StringAttr::get(context, mlir::spirv::getEntryPointABIAttrName());
-    auto abi = mlir::spirv::getEntryPointABIAttr(llvm::None, context);
-    for (auto gpuFunc : gpuModule.getOps<mlir::gpu::GPUFuncOp>()) {
-      if (!mlir::gpu::GPUDialect::isKernel(gpuFunc) ||
-          gpuFunc->getAttr(attrName))
-        continue;
+    if (m_clientAPI == "opencl") {
```
drprajap, line 58: it seems only getEntryPointABIAttr argument differs between opencl and vulkan, can this code be restructured to avoid code duplication?

Dewei-Wang-sh (approved):
lgtm overall

lib/Conversion/PTensorToLinalg/PTensorToLinalg.cpp
```diff
@@ -270,18 +270,17 @@ static BodyType buildTrivial(::mlir::Type typ) {
               builder.create<IOP>(loc, lhs, rhs).getResult());
         return;
       } else
-        assert("Found integer type but binary op not defined for integers" ==
-               nullptr);
+        assert(0 &&
```
Hardcode84, line 273: you can also do `assert(!"something")` or just `llvm_unreachable("something")`

Hardcode84:
Did it picked up cache from main? You could just change `LLVM_CACHE_NUMBER` to force cache rebuild

fschlimb:
> Did it picked up cache from main? You could just change `LLVM_CACHE_NUMBER` to force cache rebuild

Every branch has its own cache. So the CI will not reuse the cache of a given branch for the merge commit. 
Only the cache from main can be used on other branches. That's why it is desirable to have main use the same LLVM SHA as refactor and it's forks.

PR #425, opened 2022-10-30
Add spirv capability attribute to the gpu.module

nbpatel:

Add the spirv capabitity attribute only to the gpu.modules

### DIFF
--- a/lib/Transforms/SetSPIRVCapabilities.cpp
+++ b/lib/Transforms/SetSPIRVCapabilities.cpp
@@ -15,6 +15,8 @@
 
 #include <imex/Transforms/Passes.h>
 
+#include "mlir/Dialect/GPU/IR/GPUDialect.h"
+#include "mlir/Dialect/SPIRV/IR/SPIRVDialect.h"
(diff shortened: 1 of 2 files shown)


lib/Transforms/SetSPIRVCapabilities.cpp
```diff
@@ -53,8 +55,10 @@ struct SetSPIRVCapabilitiesPass
         triple, spirv::Vendor::Unknown, spirv::DeviceType::Unknown,
         spirv::TargetEnvAttr::kUnknownDeviceID,
         spirv::getDefaultResourceLimits(context));
-    auto module = getOperation();
-    module->setAttr(spirv::getTargetEnvAttrName(), attr);
+    auto op = getOperation();
+    op->walk([&](mlir::gpu::GPUModuleOp op) {
```
drprajap, line 59: this would walk through all gpu modules and sets spv attribute, is this the only attribute that we need to take care? how did you find this out?

nbpatel: it was complaining about it while running e2e through runner

drprajap (approved):

PR #437, opened 2022-11-03
Enable JAX tests for CPU runner and control exclusion of individual tests.

silee2:
This PR enables all JAX tests for CPU End-To-End testing except one test case
jit__get_age_weights_from_tables.8_linalg.mlir that does not converge based on the current input.
That test is excluded by using update config.excludes in lit.local.cfg and will be enabled in a separate PR.
The same mechanism can be used to control exclusion of individual test files in general.


nbpatel (approved):

Dewei-Wang-sh (approved):
lgtm
it would be good if we can have a single source of truth for the "linalg-to-cpu.pp"

silee2:
> lgtm it would be good if we can have a single source of truth for the "linalg-to-cpu.pp"

Will address in a separate PR.

PR #438, opened 2022-11-04
Add gpu f64 truncation pass

Hardcode84:
* For devices without f64 support, convert all computations to f32
* F64 arrays are also supported, but conversion between f32 and f64 values is done using emulation (denormals and nans/infs are not supported)
* TODO: add user option to control this


Hardcode84:
@diptorupd @nbpatel @silee2 FYI

PR #440, opened 2022-11-04
Check for llvm version while configuring

nbpatel:

This PR adds a check in the top level CMake to check if the llvm SHA in llvm_version.txt is same as the one which MLIR_DIR refers to.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -39,6 +39,20 @@ else()
     message(STATUS "Building IMEX with external MLIR")
     find_package(MLIR REQUIRED CONFIG)
 
+    # Check if the llvm SHA is consistent with llvm_version.txt
+    file(STRINGS ${CMAKE_CURRENT_SOURCE_DIR}/build_tools/llvm_version.txt EXPECTED_LLVM_SHA)
+    message(STATUS "Expected llvm sha: \"${EXPECTED_LLVM_SHA}\"")
+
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -55,6 +55,20 @@ include(AddLLVM)
 include(AddMLIR)
 include(HandleLLVMOptions)
 
+# Check if the llvm SHA is consistent with llvm_version.txt
```
silee2, line 58: Can you wrap the whole part inside "if(NOT IMEX_EXTERNAL_PROJECT_BUILD)"?
In tree build does not have "VCSRevision.h" and build will fail.
So it needs a different way of checking LLVM SHA. That support can be added in a separate PR once I figure out how to do it.

silee2 (changes requested):

CMakeLists.txt
Hardcode84: > That support can be added in a separate PR once I figure out how to do it.

Can we just call git directly from llvm folder in this case?

drprajap: also, as we talked offline, in case of different llvm SHA, build_imex.py currently checks out llvm SHA and builds it, if this fails earlier than that, you need a way how to bypass this failure when someone uses build_imex.py.

nbpatel: made the change

drprajap (approved):

silee2 (approved):

PR #452, opened 2022-11-08
Add explicit_padding gpu test

Hardcode84:
* Also, disable `linalg_test_f64` test as it will fail on devices without f64 support and we have f32 version


Hardcode84:
@nbpatel FYI

PR #454, opened 2022-11-08
change minimum required cmake version

drprajap:
When using Anaconda/miniconda environment, CMake Findpackage finds system python path instead of anaconda environment one, Detection of anaconda is introduced in later version of cmake (possibly >=3.17).CMake older version finds system python which tries to locate lit package in system path as well. Due to this lit tests are not able to find lit module.
  
Changing minimum cmake version results in finding intended python package from configured environment and can run lit tests successfully which used to fail on some environments.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.13...3.18)
+cmake_minimum_required(VERSION 3.13.4...3.18)
 
 # CMP0116: Ninja generators transform `DEPFILE`s from `add_custom_command()`
 # New in CMake 3.20. https://cmake.org/cmake/help/latest/policy/CMP0116.html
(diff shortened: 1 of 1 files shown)


nbpatel (approved):
LGTM

CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.13.4)
+cmake_minimum_required(VERSION 3.13...3.18)
```
silee2, line 1: This will introduce unwanted behavior change regarding cmake policies.
https://cmake.org/cmake/help/latest/command/cmake_minimum_required.html#policy-settings

silee2 (changes requested):

silee2 (approved):
LGTM.
Please change 3.13 -> 3.13.4

PR #455, opened 2022-11-09
[Transform] add outer parallel loop if necessary before gpu-map-parallel

Dewei-Wang-sh:
When the original function does not have an outer parallel loop, this pass adds one so that the immediately followed pass gpu-map-parallel-loops can work and generate the gpu.module.
this pass is strongly based on the assumption that the kernel code we want to offload has a top-level for-loop and the host code do not have a for-loop.
Alternatively, an annotated function with somewhat attribute which specifies this is a kernel function would also help.

### DIFF
--- a/include/imex/Transforms/Passes.td
+++ b/include/imex/Transforms/Passes.td
@@ -71,7 +71,6 @@ def AddOuterParallelLoop : Pass<"imex-add-outer-parallel-loop", "::mlir::func::F
   }];
   let constructor = "imex::createAddOuterParallelLoopPass()";
   let dependentDialects = [
-    "::mlir::gpu::GPUDialect",
     "::mlir::scf::SCFDialect"
     ];
 }
--- a/include/imex/Transforms/Passes.td
+++ b/include/imex/Transforms/Passes.td
@@ -63,4 +63,17 @@ def SetSPIRVAbiAttribute : Pass<"set-spirv-abi-attrs", "::mlir::gpu::GPUModuleOp
   ];
 }
 
+def AddOuterParallelLoop : Pass<"imex-add-outer-parallel-loop", "::mlir::func::FuncOp"> {
+  let summary = "add an outer parallel loop when there is not";
+  let description = [{
+    When the original func does not have an outer parallel loop, this pass adds
+    one so that the immediately followed pass gpu-map-parallel-loops can work.
+  }];
+  let constructor = "imex::createAddOuterParallelLoopPass()";
+  let dependentDialects = [
+    "::mlir::gpu::GPUDialect",
+    "::mlir::scf::SCFDialect"
+    ];
+}
+
 #endif // _IMEX_TRANSFORMS_PASSES_TD_INCLUDED_
--- a/lib/Transforms/AddOuterParallelLoop.cpp
+++ b/lib/Transforms/AddOuterParallelLoop.cpp
@@ -1,0 +1,58 @@
+//===- AddOuterParallelLoop.cpp - add outer parallel loop pass--*- C++ -*-===//
+//
+// Copyright 2022 Intel Corporation
+// Part of the IMEX Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+///
+/// \file
+/// When the original func does not have an outer parallel loop, this pass adds
+/// one so that the immediately followed pass gpu-map-parallel-loops can work.
+///
+//===----------------------------------------------------------------------===//
+
+#include "PassDetail.h"
+#include "mlir/Dialect/Arith/IR/Arith.h"
+#include "mlir/Dialect/GPU/IR/GPUDialect.h"
+#include "mlir/Dialect/SCF/IR/SCF.h"
+
+using namespace mlir;
+using namespace imex;
+
+namespace {
+struct AddOuterParallelLoopPass
+    : public AddOuterParallelLoopBase<AddOuterParallelLoopPass> {
+public:
+  void runOnOperation() override {
+    auto func = getOperation();
+    llvm::SmallVector<scf::ForOp, 4> ops;
+    func.walk<WalkOrder::PreOrder>([&](scf::ForOp op) {
+      if (op->getParentOp() == func) {
+        ops.push_back(op);
+        return WalkResult::skip();
+      }
+      return WalkResult::advance();
+    });
+    mlir::OpBuilder builder(func.getContext());
+    for (scf::ForOp op : ops) {
+      builder.setInsertionPoint(op);
+      mlir::Value cst0 = builder.create<arith::ConstantOp>(
+          op.getLoc(), builder.getIndexAttr(0));
+      mlir::Value cst1 = builder.create<arith::ConstantOp>(
+          op.getLoc(), builder.getIndexAttr(1));
+      scf::ParallelOp outer = builder.create<scf::ParallelOp>(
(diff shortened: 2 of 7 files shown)


fschlimb:
LGTM.

Am I right that this creates a loop with a single operation? The nature of the created loops could be mentioned somewhere.

Some comment(s) in the actual code would be nice ( `// find all scf::ForOps` ... `// add outer loops`)

Dewei-Wang-sh:
> LGTM.
> 
> Am I right that this creates a loop with a single operation? The nature of the created loops could be mentioned somewhere.
> 
> Some comment(s) in the actual code would be nice ( `// find all scf::ForOps` ... `// add outer loops`)

yes, it is a scf ParallelOp with a single original scf ForOp inside.
and add more comments

lib/Transforms/AddOuterParallelLoop.cpp
```diff
@@ -0,0 +1,60 @@
...
+    mlir::OpBuilder builder(func.getContext());
+    // move the for-loop inside the newly created parallel-loop
+    for (scf::ForOp op : ops) {
+      builder.setInsertionPoint(op);
+      mlir::Value cst0 = builder.create<arith::ConstantOp>(
+          op.getLoc(), builder.getIndexAttr(0));
+      mlir::Value cst1 = builder.create<arith::ConstantOp>(
+          op.getLoc(), builder.getIndexAttr(1));
+      scf::ParallelOp outer = builder.create<scf::ParallelOp>(
+          op.getLoc(), SmallVector<Value, 2>{cst0}, SmallVector<Value, 2>{cst1},
```
Hardcode84, line 48: nit: `Location loc = op.getLoc();` instead of calling `getLoc()` multiple times
nit: I think `cst0` should work just fine instead of `SmallVector<Value, 2>{cst0}`

include/imex/Transforms/Passes.td
```diff
@@ -63,4 +63,17 @@ def SetSPIRVAbiAttribute : Pass<"set-spirv-abi-attrs", "::mlir::gpu::GPUModuleOp
...
 
+def AddOuterParallelLoop : Pass<"imex-add-outer-parallel-loop", "::mlir::func::FuncOp"> {
+  let summary = "add an outer parallel loop when there is not";
+  let description = [{
+    When the original func does not have an outer parallel loop, this pass adds
+    one so that the immediately followed pass gpu-map-parallel-loops can work.
+  }];
+  let constructor = "imex::createAddOuterParallelLoopPass()";
+  let dependentDialects = [
+    "::mlir::gpu::GPUDialect",
```
Hardcode84, line 74: you don't actually use any ops from gpu dialect, so this can be removed

lib/Transforms/AddOuterParallelLoop.cpp
```diff
@@ -0,0 +1,60 @@
...
+    });
+    mlir::OpBuilder builder(func.getContext());
+    // move the for-loop inside the newly created parallel-loop
+    for (scf::ForOp op : ops) {
+      builder.setInsertionPoint(op);
+      mlir::Value cst0 = builder.create<arith::ConstantOp>(
+          op.getLoc(), builder.getIndexAttr(0));
+      mlir::Value cst1 = builder.create<arith::ConstantOp>(
+          op.getLoc(), builder.getIndexAttr(1));
+      scf::ParallelOp outer = builder.create<scf::ParallelOp>(
```
Hardcode84, line 47: This code doesn't support `ForOp` with reductions (it returns a value in this case), please at least check for it and add test

test/Transforms/add-outer-parallel-loop.mlir
```diff
@@ -0,0 +1,22 @@
...
+  %c0 = arith.constant 0 : index
+  %c10 = arith.constant 10 : index
+  %c1 = arith.constant 1 : index
+  %c20 = arith.constant 20 : index
+  %cst = arith.constant 0.000000e+00 : f32
+  %alloc = memref.alloc() {alignment = 128 : i64} : memref<f32>
+  memref.store %cst, %alloc[] : memref<f32>
+  %alloc_0 = memref.alloc() {alignment = 128 : i64} : memref<f32>
+  memref.copy %alloc, %alloc_0 : memref<f32> to memref<f32>
+  // CHECK: scf.parallel
```
Hardcode84, line 12: please check for loops are actually inside this `scf.parallel`

test/Transforms/add-outer-parallel-loop.mlir
```diff
@@ -0,0 +1,22 @@
+// RUN: imex-opt --imex-add-outer-parallel-loop %s | FileCheck %s
+func.func @test(%arg0: memref<10x20xf32>) -> memref<f32> {
```
Hardcode84, line 2: `// CHECK_LABEL: func @test `

test/Transforms/add-outer-parallel-loop.mlir
```diff
@@ -0,0 +1,22 @@
...
+  scf.for %arg1 = %c0 to %c10 step %c1 {
+    scf.for %arg2 = %c0 to %c20 step %c1 {
+      %0 = memref.load %arg0[%arg1, %arg2] : memref<10x20xf32>
+      %1 = memref.load %alloc_0[] : memref<f32>
+      %2 = arith.addf %1, %0 : f32
+      memref.store %2, %alloc_0[] : memref<f32>
+    }
+  }
+  return %alloc_0 : memref<f32>
+}
```
Hardcode84, line 22: please add test for 2 consecutive `scf.for`'s

Dewei-Wang-sh: added

test/Transforms/add-outer-parallel-loop.mlir
Dewei-Wang-sh: added

lib/Transforms/AddOuterParallelLoop.cpp
Dewei-Wang-sh: yes, it works, thanks for the detailed review.

include/imex/Transforms/Passes.td
Dewei-Wang-sh: removed

test/Transforms/add-outer-parallel-loop.mlir
Dewei-Wang-sh: added

lib/Transforms/AddOuterParallelLoop.cpp
```diff
@@ -0,0 +1,60 @@
...
+    // populate the top level for-loop
+    func.walk<WalkOrder::PreOrder>([&](scf::ForOp op) {
+      if (op->getParentOp() == func) {
+        ops.push_back(op);
+        return WalkResult::skip();
+      }
+      return WalkResult::advance();
+    });
+    mlir::OpBuilder builder(func.getContext());
+    // move the for-loop inside the newly created parallel-loop
```
nbpatel, line 40: where do you check if there is no outer scf.parallel  loop?

Dewei-Wang-sh: in the code above, I only check the forOp whose parentOp is the FuncOp.
so if there is an outer scf.parallel, it wouldn't be the case.

lib/Transforms/AddOuterParallelLoop.cpp
Dewei-Wang-sh: added a straightforward way to handle ForOp with iter_args(move all its users into the parallel loop)

test/Transforms/add-outer-parallel-loop.mlir
Hardcode84: nit: you can improve this check by also matching closing `}`

Hardcode84 (approved):

nbpatel (approved):
LGTM

PR #507, opened 2022-11-21
[Transforms] lower memref.copy

Dewei-Wang-sh:
add an optional pass to lower memref.copy  after bufferization and before linalg-lowering.
with this, we can open more chance to later optimization like loop-fusion/store-load-forwarding.
also it make sense to process load/store in a memref.copy  in the same way with load/store in a linalg.generic op.

### DIFF
--- a/lib/Transforms/LowerMemRefCopy.cpp
+++ b/lib/Transforms/LowerMemRefCopy.cpp
@@ -1,5 +1,4 @@
-//===- LowerMemRefCopy.cpp - lower memref.copy pass --------------*- C++
-//-*-===//
+//===- LowerMemRefCopy.cpp - lower memref.copy pass --------*- C++ -*-===//
 //
 // Copyright 2022 Intel Corporation
 // Part of the IMEX Project, under the Apache License v2.0 with LLVM Exceptions.
@@ -45,14 +44,14 @@ struct LowerMemRefCopy
             return WalkResult::skip();
           auto src = op.getSource();
           auto dst = op.getTarget();
-          // supposed to work on memref type
+          // supposed to work on same memref type
           auto srcType = src.getType().cast<MemRefType>();
           auto dstType = dst.getType().cast<MemRefType>();
-          if (!srcType || !dstType || srcType != dstType)
+          if (srcType != dstType)
             return WalkResult::skip();
           // supposed to work on memref.alloc
-          auto srcOp = cast<memref::AllocOp>(src.getDefiningOp());
-          auto dstOp = cast<memref::AllocOp>(dst.getDefiningOp());
+          auto srcOp = src.getDefiningOp<memref::AllocOp>();
+          auto dstOp = dst.getDefiningOp<memref::AllocOp>();
           if (!srcOp || !dstOp)
             return WalkResult::skip();
           // check use of src after this copyOp, being conservative
@@ -72,12 +71,11 @@ struct LowerMemRefCopy
           if (hasSubsequentUse) {
             OpBuilder builder(op);
             linalg::makeMemRefCopyOp(builder, op.getLoc(), src, dst);
-            op.erase();
           } else {
             // coalesce buffer
             dst.replaceAllUsesWith(src);
-            op.erase();
           }
+          op.erase();
           return WalkResult::advance();
         });
   }
--- a/lib/Transforms/LowerMemRefCopy.cpp
+++ b/lib/Transforms/LowerMemRefCopy.cpp
@@ -1,0 +1,91 @@
+//===- LowerMemRefCopy.cpp - lower memref.copy pass --------------*- C++
+//-*-===//
+//
+// Copyright 2022 Intel Corporation
+// Part of the IMEX Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+///
+/// \file
+/// This pass lowers memref copyOp to linalg generic operations and enables
+/// simple memref copyOp canonicalization
+///
+//===----------------------------------------------------------------------===//
+
+#include "imex/Transforms/Passes.h"
+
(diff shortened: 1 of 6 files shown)


Hardcode84:
Why don't just use `makeMemRefCopyOp` https://github.com/llvm/llvm-project/blob/main/mlir/include/mlir/Dialect/Linalg/Utils/Utils.h#L129 ?

Hardcode84:
And, in general, you are trying to do 2 different things here:
* Remove unnessary `memref.copy`s
* Convert remaining `memref.copy` to linalg

I suggest to split them into 2 separate passes, where second one will be just trivial call to `makeMemRefCopyOp`.

nbpatel:
why do we need this?

nbpatel:
is this for the memRefCopy symbol not found issue? if yes, thats because we dont include c_runner_util library....if you just pass that as a part of shared. lib it will work.....
https://github.com/llvm/llvm-project/blob/894f9e54f59522f57b7a106b0c2f031260964ce8/mlir/lib/ExecutionEngine/CRunnerUtils.cpp#L49

Hardcode84:
> why do we need this?

This is actually make sense from perf point of view, because those `linalg.generic`s can be potentially fused with something else (but you don't have any fusion passes in your pipeline yet, though)

nbpatel:
> > why do we need this?
> 
> This is actually make sense from perf point of view, because those `linalg.generic`s can be potentially fused with something else (but you don't have any fusion passes in your pipeline yet, though)

we haven't started any performance work yet. but ya, we might need it in future, I guess the purpose of it would be what I mentioned above

Dewei-Wang-sh:
> > > why do we need this?
> > 
> > 
> > This is actually make sense from perf point of view, because those `linalg.generic`s can be potentially fused with something else (but you don't have any fusion passes in your pipeline yet, though)
> 
> we haven't started any performance work yet. but ya, we might need it in future, I guess the purpose of it would be what I mentioned above

yes, here I look backwards and found that the memcopy can be handled in a early place.
I will make this optional and just use the c_runner_utils to make it work

Dewei-Wang-sh:
> Why don't just use `makeMemRefCopyOp` https://github.com/llvm/llvm-project/blob/main/mlir/include/mlir/Dialect/Linalg/Utils/Utils.h#L129 ?

thanks, I will call the existing func to ease the effort

Dewei-Wang-sh:
> And, in general, you are trying to do 2 different things here:
> 
> * Remove unnessary `memref.copy`s
> * Convert remaining `memref.copy` to linalg
> 
> I suggest to split them into 2 separate passes, where second one will be just trivial call to `makeMemRefCopyOp`.

I agree, there 2 things are all related to memref copyOp, so I put it all in this pass.
since we are in a standalone repo, otherwise I would add a canonicalize pattern to memref.
will move "Remove unnessary `memref.copy`s" part to a seperate place later

lib/Transforms/LowerMemRefCopy.cpp
```diff
@@ -0,0 +1,91 @@
...
+            return WalkResult::skip();
+          auto src = op.getSource();
+          auto dst = op.getTarget();
+          // supposed to work on memref type
+          auto srcType = src.getType().cast<MemRefType>();
+          auto dstType = dst.getType().cast<MemRefType>();
+          if (!srcType || !dstType || srcType != dstType)
+            return WalkResult::skip();
+          // supposed to work on memref.alloc
+          auto srcOp = cast<memref::AllocOp>(src.getDefiningOp());
```
Hardcode84, line 54: `getDefiningOp<memref::AllocOp>()`

lib/Transforms/LowerMemRefCopy.cpp
```diff
@@ -0,0 +1,91 @@
...
+    auto func = getOperation();
+    // walk through memref.copy ops in the funcion body
+    WalkResult result =
+        func.walk<WalkOrder::PreOrder>([&](memref::CopyOp op) -> WalkResult {
+          if (op->getParentOp() != func)
+            return WalkResult::skip();
+          auto src = op.getSource();
+          auto dst = op.getTarget();
+          // supposed to work on memref type
+          auto srcType = src.getType().cast<MemRefType>();
```
Hardcode84, line 49: `cast` will assert if it isn't possible, you need `dyn_cast`

lib/Transforms/LowerMemRefCopy.cpp
```diff
@@ -0,0 +1,91 @@
...
+              hasSubsequentUse = true;
+              break;
+            }
+          }
+
+          // replace copy with linalg.generic
+          if (hasSubsequentUse) {
+            OpBuilder builder(op);
+            linalg::makeMemRefCopyOp(builder, op.getLoc(), src, dst);
+            op.erase();
```
Hardcode84, line 75: nit: `op.erase()` can be moved after if

Dewei-Wang-sh: changed

lib/Transforms/LowerMemRefCopy.cpp
Dewei-Wang-sh: my bad, the td definition ensures that it is memrefType, so eliminate the null check.

lib/Transforms/LowerMemRefCopy.cpp
Dewei-Wang-sh: changed

lib/Transforms/LowerMemRefCopy.cpp
```diff
@@ -0,0 +1,90 @@
+//===- LowerMemRefCopy.cpp - lower memref.copy pass --------------*- C++
+//-*-===//
```
Hardcode84, line 2: broken formatting?

Hardcode84 (approved):
LGTM

lib/Transforms/LowerMemRefCopy.cpp
Dewei-Wang-sh: yes, seems pre-commit changed this.

silee2 (approved):
LGTM

PR #536, opened 2022-11-28
Enable I8 and I1 testcases

chencha3:
Enable a few more PlaidML test cases, for those failed are now temporarily excluded in lit.local.config


Dewei-Wang-sh (approved):
lgtm

nbpatel (approved):

PR #538, opened 2022-11-30
Change imex-runner.py pass pipeline file comment from Python style

silee2:
comment to C++ style comment


silee2:
pass pipeline lacks documentation on syntax. Documentation will be added in a separate PR. Created an issue to track it. #539

mshahneo (approved):
LGTM

nbpatel (commented):
Just one point...usually dont the python files use # for comments? So I thought # made more sense in this case...// is usually used for .c, .cpp files... maybe I am wrong

silee2:
> Just one point...usually dont the python files use # for comments? So I thought # made more sense in this case...// is usually used for .c, .cpp files... maybe I am wrong

@nbpatel Yes "#" makes more sense for comment. And that is why I used "#" originally. But mlir pass options include using "#" as a character so had to pick something else for comment.

nbpatel (approved):
LGTM

PR #545, opened 2022-12-02
Reductions support for GPU

Hardcode84:
Some background: upstream gpu lowering doesn't support lowering `scf.parallel` ops with reductions to the gpu ops, so we need some additional transformations before running it.
1. Convert `scf.parallel` reduction to our custom `gpu_runtime.global_reduce` op (was done in separate PR)
2. Lower  `gpu_runtime.global_reduce` to sequence of actual gpu ops
* There are multiple ways to do a reduction on GPU but most of them involves doing reduction across workgroup and then aggregating final result in some way
* Current algorithm is 
    * Use dedicated spirv group ops to do a WG reduction and store result to global shared buffer
    * Then aggregate results from buffer on host
 * This final host aggregation step can be avoided by using spirv atomic ops or repeatedly calling the reduction kernel but performance implications of either of those options are TBD
 * Test cases for 1-sized arrays are disabled as the compiler is too smart in this case and manages to remove reduction loop altogether, leaving single read from src array instead of reduction, which is crashes as we trying to read gpu memory on host.

### DIFF
--- a/mlir/include/imex/Conversion/GpuToGpuRuntime.hpp
+++ b/mlir/include/imex/Conversion/GpuToGpuRuntime.hpp
@@ -42,4 +42,8 @@ std::unique_ptr<mlir::Pass> createTruncateF64ForGPUPass();
 /// This pass is intended to be run right before scf-to-gpu.
 std::unique_ptr<mlir::Pass> createInsertGPUGlobalReducePass();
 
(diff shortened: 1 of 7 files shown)


Hardcode84:
@diptorupd @nbpatel FYI

PR #548, opened 2022-12-04
[Profiling] add profiling code to get kernel submit/execute time

Dewei-Wang-sh:
take test/PlaidML/OpTest.EltwiseAdd.mlir for example, we can get the following result.
the kernel execution time is 0.082560 ms
the kernel submission time is 97647.601562 ms  
( this duration is obviously wrong cause the test is done within 3 seconds, still trying to find the reason)

### DIFF
--- a/include/imex/Transforms/Passes.td
+++ b/include/imex/Transforms/Passes.td
@@ -46,6 +46,7 @@ def InsertGPUAllocs : Pass<"insert-gpu-allocs", "::mlir::func::FuncOp"> {
 def SetSPIRVCapabilities : Pass<"set-spirv-capabilities"> {
   let summary = "Sets Spirv capabilities";
   let constructor = "imex::createSetSPIRVCapabilitiesPass()";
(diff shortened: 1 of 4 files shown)


include/imex/Transforms/Passes.td
```diff
@@ -46,6 +46,7 @@ def InsertGPUAllocs : Pass<"insert-gpu-allocs", "::mlir::func::FuncOp"> {
 def SetSPIRVCapabilities : Pass<"set-spirv-capabilities"> {
   let summary = "Sets Spirv capabilities";
   let constructor = "imex::createSetSPIRVCapabilitiesPass()";
+  let dependentDialects = ["::mlir::spirv::SPIRVDialect"];
```
Dewei-Wang-sh, line 49: without this, it will break when run this pass alone.

silee2 (approved):
LGTM

nbpatel:
Thanks for this PR.
Can we add something similar for L0 as well?

chencha3 (approved):
LGTM

PR #553, opened 2022-12-06
Fix IMEX config and installation issues.

silee2:
Fix issues with IMEXConfig.cmake and IMEXTargets.cmake and installation issues.
Should be able to replace #290
Some imported cmake functions(add_mlir_* ) are redefined for the fix.
Added new macro add_mlir_tool() that is used for handling imex-opt installation.
Added an example CMake project using external IMEX under "integration/example"

Also includes some changes to 
- header inclusion
- install rules
- cleanup cmake file (remove unused variables)

Please review these guidelines to help with the review process:
- [ ] Have you provided a meaningful PR description?
- [ ] Have you added a test, a reproducer, or a reference to an issue with a reproducer?
- [ ] Have you tested your changes locally for CPU and GPU devices?
- [ ] Have you made sure that new changes do not introduce compiler warnings?
- [ ] If this PR is a work in progress, are you filing the PR as a draft?
- [ ] Have you organized your commits logically and ensured each can be built by itself?

PR #554, opened 2022-12-06
[Format] change header include style to quotes

Dewei-Wang-sh:
change all imex and llvm include style from brackets to quotes

### DIFF
--- a/lib/Conversion/DistToStandard/DistToStandard.cpp
+++ b/lib/Conversion/DistToStandard/DistToStandard.cpp
@@ -13,18 +13,18 @@
 ///
 //===----------------------------------------------------------------------===//
 
-#include <imex/Conversion/DistToStandard/DistToStandard.h>
-#include <imex/Dialect/Dist/IR/DistOps.h>
-#include <imex/Dialect/PTensor/IR/PTensorOps.h>
-#include <imex/Utils/PassUtils.h>
-#include <imex/Utils/PassWrapper.h>
+#include "imex/Conversion/DistToStandard/DistToStandard.h"
+#include "imex/Dialect/Dist/IR/DistOps.h"
+#include "imex/Dialect/PTensor/IR/PTensorOps.h"
+#include "imex/Utils/PassUtils.h"
+#include "imex/Utils/PassWrapper.h"
 
-#include <mlir/Dialect/Arith/IR/Arith.h>
-#include <mlir/Dialect/Func/IR/FuncOps.h>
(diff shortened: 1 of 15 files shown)


fschlimb (changes requested):
I don't think we ever want anything else than including the files from an install-like structure. So why are you changing the include style? ""s don't make sense to me.

silee2:
> I don't think we ever want anything else than including the files from an install-like structure. So why are you changing the include style? ""s don't make sense to me.

This is a quick fix for an external customer but I agree that this not the right solution.

imex headers are local so I think using quotes to include makes sense.
Meanwhile, mlir and llvm headers are external so using brackets to include makes sense.

So I think IMEX implementation files should include headers like the following.

 #include <mlir/...>
 #include <llvm/...>
 #include "imex/..."

silee2:
#553 adds llvm and mlir headers with "include_directories(SYSTEM ...)"  and imex headers with "include_directories(...)"

fschlimb:
> imex headers are local so I think using quotes to include makes sense.

Quotes make sense only 
- for including files which are somewhere within the directory of the current file. The only benefit this can bring is that paths in the include directive can be shorter.
- you want to allow replacing "system" include files by providing alternatives within the same directory

The latter is dangerous and should be handled differently (by adjusting include paths).

Hence, unless including a header from within the same directory, it is more robust to use brackets no matter if including from imex or not.

silee2:
> > imex headers are local so I think using quotes to include makes sense.
> 
> Quotes make sense only
> 
> * for including files which are somewhere within the directory of the current file. The only benefit this can bring is that paths in the include directive can be shorter.
> * you want to allow replacing "system" include files by providing alternatives within the same directory
> 
> The latter is dangerous and should be handled differently (by adjusting include paths).
> 
> Hence, unless including a header from within the same directory, it is more robust to use brackets no matter if including from imex or not.

Makes sense.

Dewei-Wang-sh:
> > imex headers are local so I think using quotes to include makes sense.
> 
> Quotes make sense only
> 
> * for including files which are somewhere within the directory of the current file. The only benefit this can bring is that paths in the include directive can be shorter.
> * you want to allow replacing "system" include files by providing alternatives within the same directory
> 
> The latter is dangerous and should be handled differently (by adjusting include paths).
> 
> Hence, unless including a header from within the same directory, it is more robust to use brackets no matter if including from imex or not.

thanks frank and sang lk, got the difference, will close the pr and dive more

PR #563, opened 2022-12-12
Linalg matmul converter 1

shssf:
Hello, this is my first PR into this repository.
This PR is more like a question "is this right way to do the task?". It contains no code to do any transformations.

Provided a skeleton for `Conversion` of `linalg.matmul` to a `SPIRV`

I didn't prepare a test case yet because it is not clear "what should be done" in this `Converter`.


mshahneo (approved):
LGTM, let's forget about the CI for now. If it's passing the build in your system, let's move on.

PR #568, opened 2022-12-13
[Minor Fix] change default client api from vulkan to opencl

Dewei-Wang-sh:

### DIFF
--- a/include/imex/Transforms/Passes.td
+++ b/include/imex/Transforms/Passes.td
@@ -38,7 +38,7 @@ def InsertGPUAllocs : Pass<"insert-gpu-allocs", "::mlir::func::FuncOp"> {
                            "::mlir::gpu::GPUDialect",
                            "::mlir::arith::ArithDialect"];
   let options = [
-    Option<"clientAPI", "client-api", "std::string", /*default=*/"\"vulkan\"",
+    Option<"clientAPI", "client-api", "std::string", /*default=*/"\"opencl\"",
            "The client API to use for inserting gpu allocs">
   ];
 }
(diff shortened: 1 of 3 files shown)


lib/Transforms/SetSPIRVCapabilities.cpp
```diff
@@ -111,7 +110,11 @@ struct SetSPIRVCapabilitiesPass
 } // namespace
 
 namespace imex {
+std::unique_ptr<mlir::Pass>
+createSetSPIRVCapabilitiesPass(mlir::StringRef api) {
+  return std::make_unique<SetSPIRVCapabilitiesPass>(api);
```
Dewei-Wang-sh, line 115: add this new ctor so that user explicitly calling this createXXX method can pass parameter.

test/PlaidML/OpTest.Conv1D.mlir
```diff
@@ -54,7 +54,7 @@ module @conv_1d {
   func.func private @printMemrefF32(%ptr : tensor<*xf32>)
 }
 //      CHECK: Unranked Memref base@ = {{(0x)?[-9a-f]*}}
-// CHECK-NEXT: 76
+// CHECK-NEXT: [76
```
Dewei-Wang-sh, line 57: in case that the above memref base can have 76

include/imex/Transforms/Passes.td
```diff
@@ -38,7 +38,7 @@ def InsertGPUAllocs : Pass<"insert-gpu-allocs", "::mlir::func::FuncOp"> {
                            "::mlir::gpu::GPUDialect",
                            "::mlir::arith::ArithDialect"];
   let options = [
-    Option<"clientAPI", "client-api", "std::string", /*default=*/"\"vulkan\"",
+    Option<"clientAPI", "client-api", "std::string", /*default=*/"\"opencl\"",
```
Dewei-Wang-sh, line 41: since our end to end pipeline use opencl, so change the default here

nbpatel (approved):

PR #594, opened 2023-01-05
Fix checkmarx errors

nbpatel:

This PR fixes the checkmarx "high" vulnerabilities issues


drprajap:
@nbpatel, if there is any way to add this build_imex script to ignore list, it is better to do it, we should focus only on IMEX source owned by us (even ignore llvm headers if it is flagged)

PR #597, opened 2023-01-10
Importing Fusion Pass

chencha3:
importing linalg fusion pass. It is not the final solution yet, but it can help to fuse some simple loops.


Dewei-Wang-sh (approved):
lgtm

PR #602, opened 2023-01-11
Linalg matmul converter test

shssf:
Adding a test case with reference to allow running and compre results


mshahneo (approved):
Looks good to me.
@shssf, which machine did you test it on?

shssf:
@mshahneo I used CPU based machine (local one). This is my question how I can use GPU to run the code. We could discuss it on a meeting.

PR #609, opened 2023-01-12
[coverity issue] fix uninitialized var

Dewei-Wang-sh:

### DIFF
--- a/lib/Transforms/AddOuterParallelLoop.cpp
+++ b/lib/Transforms/AddOuterParallelLoop.cpp
@@ -31,7 +31,7 @@ public:
     auto func = getOperation();
     if (func.getBody().empty())
       return;
-    std::vector<llvm::SmallVector<Operation *, 4>> groupedOps;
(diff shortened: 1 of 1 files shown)


lib/Transforms/AddOuterParallelLoop.cpp
```diff
@@ -31,7 +31,7 @@ struct AddOuterParallelLoopPass
     auto func = getOperation();
     if (func.getBody().empty())
       return;
-    std::vector<llvm::SmallVector<Operation *, 4>> groupedOps;
+    llvm::SmallVector<llvm::SmallVector<Operation *, 4>, 4> groupedOps;
```
drprajap, line 34: Is "4" here specifies value? If So, is there specific reason to initialize with 4?

Dewei-Wang-sh: it's a common practice in llvm to specify it as 4 or 8 according to the using case.

drprajap (approved):

PR #679, opened 2024-01-26
[WIP] Some support for 'order' attribute in XeTile -> XeGpu lowering

Hardcode84:
* Supported in `InitTile` and `LoadTile`
* Not clear how to support in `StoreTile` as it missing transpose attr
* FIXME: Disabled encoding verification in `LoadTileOp` for now as you have to use it if you want to have order/transpose and tiling pass does not set `innerBlocks` properly.


lib/Dialect/XeTile/IR/XeTileOps.cpp
```diff
@@ -351,7 +351,8 @@ bool verifyInnerBlocksWithVecShape(mlir::DenseI32ArrayAttr &innerBlocks,
 mlir::LogicalResult LoadTileOp::verify() {
   auto encoding = getSource().getType().getEncoding();
   auto tileShape = getSource().getType().getShape();
-  if (!encoding)
+  // FIXME: TileAttr innerBlocks is not being setup properly by tiling pass
```
charithaintc, line 354: what would be the fix for this? I believe tiling pass needs to  attach this info into tile data type?

test/Conversion/XeTileToXeGPU/xetile_to_xegpu_order.mlir
```diff
@@ -0,0 +1,77 @@
...
+//       CHECK: %[[R1:.*]] = xegpu.load_nd %[[T1]] {mode = vc, transpose = [1, 0], l1_hint = cached, l2_hint = cached, l3_hint = cached} : !xegpu.tensor_desc<8x16xf16> -> vector<16x8xf16>
+//       CHECK: %[[R2:.*]] = xegpu.load_nd %[[T2]] {mode = vc, transpose = [1, 0], l1_hint = cached, l2_hint = cached, l3_hint = cached} : !xegpu.tensor_desc<8x16xf16> -> vector<16x8xf16>
+//       CHECK: %[[R3:.*]] = xegpu.load_nd %[[T3]] {mode = vc, transpose = [1, 0], l1_hint = cached, l2_hint = cached, l3_hint = cached} : !xegpu.tensor_desc<8x16xf16> -> vector<16x8xf16>
+//       CHECK: %{{.*}} = builtin.unrealized_conversion_cast %[[R0]], %[[R1]], %[[R2]], %[[R3]]
+func.func @test_init_tile(%A: memref<128x64xf16>) {
+    %c0 = arith.constant 0 : index
+    %c32 = arith.constant 32 : index
+    %c16  = arith.constant 16 : index
+    %tile = xetile.init_tile %A[%c0, %c0] : memref<128x64xf16> -> !xetile.tile<32x16xf16, #xetile.tile_attr<order = [0, 1]>>
+    %res = xetile.load_tile %tile : !xetile.tile<32x16xf16, #xetile.tile_attr<order = [0, 1]>> -> vector<16x32xf16>
```
charithaintc, line 74: I think with order attribute the output vector is not transposed. order just says the `row-major` or `col-major` property of the input mem descriptor i.e. it only specify the indexing order of the tile  access. 
https://github.com/intel/mlir-extensions/blob/main/docs/rfcs/XeTile.md

Also, I think `xetile.load_tile` is missing some verification for this case.

charithaintc:
we normally do the PRs against innersource first and then merge with public IMEX. Any reason for the exception here?

PR #681, opened 2024-02-16
Minor wording correction for order attribute

Jianhui-Li:

### DIFF
--- a/docs/rfcs/XeTile.md
+++ b/docs/rfcs/XeTile.md
@@ -79,7 +79,7 @@ Attribute `padding` specifies the padding value for the out-of-boundary access. 
 ```
 `load_tile` needs to be used with the tile_mma.
 
-`load_tile` loads a tile according to the tile's `order` attribute. Regardless of the `order` attribute value, the vector's dimensions must match exactly the Tile's dimensions. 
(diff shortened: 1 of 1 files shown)


docs/rfcs/XeTile.md
```diff
@@ -79,37 +79,37 @@ Attribute `padding` specifies the padding value for the out-of-boundary access.
 ```
 `load_tile` needs to be used with the tile_mma.
 
-`load_tile` loads a tile with an `order` attribute. The `order` attribute affects the indexing order for the tile access, however, it doesn't change the logical representation of loading a 2D tile to a 2D vector. There is no need to reverse the order of vector length at the XeTile level regardless of the `order` attribute value.
+`load_tile` loads a tile according to the tile's `order` attribute. Regardless of the `order` attribute value, the vector's dimensions must match exactly the Title's dimensions.
```
Garra1980, line 82: Title's->Tile's

chencha3: How about explicitly stating that order attribute reveals whether the base memref is in row-major or col-major order. It is only used to guide compiler for code generation and does not impact users' algorithm expression?

Jianhui-Li: fixed both comments.  
I added the description in the init_tile, "The order must be consistent with the data layout specified by the memref represening the base matirx.".

Garra1980 (approved):

PR #701, opened 2024-03-25
Add capability to attach appropriate inner block size attribute

nbpatel:
This PR adds appropriate inner_block_size attribute to init/load/store tile op based on the hardware configuration.


Garra1980 (approved):
LGTM

PR #723, opened 2024-04-18
Add pipeline composition RFC

Hardcode84:


docs/rfcs/PipelineComposition.md
```diff
@@ -0,0 +1,73 @@
...
+They can create a two empty pipelines `high-to-middle` and `midlle-to-low` with appropriate dependencies and the use those as anchors to specify at which compiler stage insert stages which do actual work.
+
+### PipelineSchedule
+```C++
+class PipelineSchedule {
+public:
+    LogicalResult run(Operation *op);
+};
+```
+`PipelineSchedule` object encapsulates compiled pipeline graph. Main method is `LogicalResult run(Operation *op);` which follows existing MLIR `PassManager::run`.
```
chencha3, line 57: What do you mean by the compiled pipeline graph?

docs/rfcs/PipelineComposition.md
```diff
@@ -0,0 +1,73 @@
...
+### PipelineGraph
+```C++
+class PipelineGraph {
+public:
+    void registerPipeline(
+        StringRef name,
+        ArrayRef<StringRef> predecessors,
+        ArrayRef<StringRef> successors,
+        ArrayRef<StringRef> jumpTargets,
+        std::function<void(OpPassManager &)> populateFunc);
```
chencha3, line 25: So a pipeline is a set of Patterns populated in populateFunc?

Hardcode84: Pipeline is set of passes.

docs/rfcs/PipelineComposition.md
Hardcode84: `PipelineGraph` object is populated by set of pipelines with dependencies, and and then it compiles them into some internal representation which runs those pipelines in order, according to those dependencies.

docs/rfcs/PipelineComposition.md
```diff
@@ -0,0 +1,73 @@
...
+```
+`PipelineGraph` is main entry point for this new API.
+User is adding pipelines into graph via `registerPipeline` function, each pipeline have the following properties:
+* `name` - Name of thepipeline
+* `predecessors` - List of names of pipelines which must be run before current pipeline.
+* `successors` - List of the names of pipelines which must be run after current pipeline.
+* `jumpTargets` - List of the names of pipelines to which we can dynamically jump after current pipeline.
+* `populateFunc` - Callback to populate pipeline with passes.
+
+After user populated the graph object they must call `createPipelineSchedule` metdod to compile the resulted graph into runnable schedule.
```
Jianhui-Li, line 38: metdod -> method

docs/rfcs/PipelineComposition.md
```diff
@@ -0,0 +1,73 @@
+# RFC: High Level MLIR Pass Pipeline Composition
+
+Ivan Butygin, Renato Golin
+
+## Motivation
+
+TBD use cases from IREE, TPP
```
Jianhui-Li, line 7: It would help reader to start with a motivation. I assume that the dependency-based graph would avoid some mistake when user configure the pipeline manually and unintentionally break the dependency. Is it correct?

docs/rfcs/PipelineComposition.md
```diff
@@ -0,0 +1,170 @@
...
+* `name` - Name of thepipeline
+* `predecessors` - List of names of pipelines which must be run before current pipeline.
+* `successors` - List of the names of pipelines which must be run after current pipeline.
+* `jumpTargets` - List of the names of pipelines to which we can dynamically jump after current pipeline.
+* `populateFunc` - Callback to populate pipeline with passes.
+
+After user populated the graph object they must call `createPipelineSchedule` method to compile the resulted graph into runnable schedule.
+`createPipelineSchedule` will build a DAG from pipelines dependencies provided by user, and will try to get linear execution order to satify these dependencies.
+
+If two pipelines doesn't have direct and indirect dependencies, order in which they will be executed is not specified, but stable.
```
rengolin, line 53: I think that's asking too much of this framework. I'd say "stability is depending on the passes accepting canonical forms from each other", and we make sure we always run canonicalization between DAG nodes.

docs/rfcs/PipelineComposition.md
rengolin: Is the schedule the result of the linearization of the DAG or is it the class that will linearize it?

docs/rfcs/PipelineComposition.md
```diff
@@ -0,0 +1,170 @@
...
+```
+builtin.module {pipeline.jump_target="Foo"} {
+...
+}
+```
+
+Passes inside pipeline can set this attribute to indicate they want compilatin flow to jump to the specific point.
+After current pipeline is finished, runtime will check if module object have attribute set and if it does, jump to the selected pipeline and clear the attribute.
+
+Setting attribute to the value, which wasnt in `jumpTargets` for the current pipeline will result in error and abort the compilation flow.
```
rengolin, line 85: `jumpTargets` seem to be used for control flow.

I'd create a conditional and looping semantics instead as a type of sub-graph.

For example a pipeline node that lists a bunch of passes (or sub-nodes) and has `arity` (ex. `until-converge-max-n`). Or another that has two sub nodes with a `select` from an IR property (ex. DLTI target information).

Giving users the ability to jump to arbitrary targets is a foot gun that we might not want to create.

docs/rfcs/PipelineComposition.md
```diff
@@ -0,0 +1,170 @@
...
+## New APIs
+
+### PipelineGraph
+```C++
+class PipelineGraph {
+public:
+    void registerPipeline(
+        StringRef name,
+        ArrayRef<StringRef> predecessors,
+        ArrayRef<StringRef> successors,
```
rengolin, line 35: I'd also avoid having both `predecessors` and `sucessors`. This feels like a duplication and hard to get right on larger graphs.

What I had in mind is just:
* Dependencies: Passes that you _must_ run before (analyses and transforms)
* Post-clean up: Canonicalization that can help the following passes

Dependencies can be bundles or specific passes. Bundles can be just a list of passes (ex. buff+ownership), a loop or a conditional (see below). Both bundles and passes have deps/cleanups and we can simplify the graph after linearization.

Post-cleanups would also be simplified (de-duped) if one pass lists it as its cleanups and the following pass lists it as its dependencies.

docs/rfcs/PipelineComposition.md
Hardcode84: Here I only meant order of pipelines/passes is stable regardless of in which order `registerPipeline`s were called (in POC impl I'm just sorting by pipeline name first to make it stable), but yes, I can remove this for more implementation freedom.

docs/rfcs/PipelineComposition.md
Hardcode84: Expanded motivation.

docs/rfcs/PipelineComposition.md
Hardcode84: `PipelineSchedule` is linearized DAG, `createPipelineSchedule` will do the linearization.

rengolin (commented):
Some items are missing from our discussion, mainly how to build the DAG and how to schedule it.

## Building the DAG

Building a DAG is simply taking all passes and insert in the first available slot (similar to tree insertion). Since there is no implicit ordering for passes, this may be restricted to `O(n^2)`.

We could reduce the complexity by creating sub-graphs inside sub-graphs and connecting the super-graphs together.

For example:
* All passes before bufferization are a sub-graph that leads into `bufferization`. There is no implicit order (needs to be scheduled). This is equivalent in saying bufferization _depends on_ all of those passes, but explicitly joining all nodes into a single one.
* Bufferization as a node with all cleanups
* Same for vectorization, lowering, etc.

```
         /----\        /----\        /----\
Ingress -------- Buff -------- Vect -------- Lower -> HW
         \----/        \----/        \----/
```

Where `Buff`, `Vect` and `Lower` are fixed sequences of passes (per target, so can be conditional).

## Scheduling

Each of those sub-graphs above will need to be scheduled. This is just graph scheduling, and can be much simpler if we hide loops and conditionals inside nodes.

Loops become a single node that is guaranteed to finish (run until convergence, but stop hard at `N` iterations, where `N` is configurable but less than a global `MaxN`).

If we follow the sub-graph design, then scheduling is always restricted to the sub-graph. This works well with a recursive algorithm that schedules the outer-most graph, then descends into sub-graphs, expanding them in linear from.

## Cleaning up

After the graph is linear (with potential loop and conditional nodes), we can start the cleanup, for example, de-duplicating passes that have no writable transforms in between.

## Failure

Failure can happen at any stage above and the error message must make clear which stage and what happened. Failed creating a DAG, sub-DAG, scheduling some sub-graph, etc.

docs/rfcs/PipelineComposition.md
Hardcode84: Regarding having both predecessors and successors,

Following (hypothetical) pipeline:
```
numpy-to-linalg  torch-to-linalg
            \      /
          bufferization
            /      \
  linalg-to-cpu  linalg-to-gpu
```
We don't want to bufferization to know about specific `***-to-linalg` pipelines, as it is a frontend details, irrelevant to bufferization, and we don't want it to know about `linalg-to-***` either as it backend details.
So pipeline should looks like
```
numpy-to-linalg: [], [bufferization]
torch-to-linalg: [], [bufferization]
bufferization: [], []
linalg-to-cpu: [bufferization], []
linalg-to-gpu: [bufferization], []
```

Hardcode84:
Subgarphs are useful by itself,  but regarding encapsulating control flow into subgraphs, let's say we have following pipeline:
```
frontend
    |
    V
python-to-standard
    |
    V
lower-to-llvm
```
```
frontend: [], [], []
python-to-standard: [frontend], [lower-to-llvm], []
lower-to-llvm: [], [], []
```
Now, (external) user wants to add `numpy-to-linalg` stage and both `python-to-standard` and `numpy-to-linalg` stages must run until fixed point.

With jumps they can just do
```
numpy-to-linalg: [python-to-standard], [lower-to-llvm], /*jump*/[python-to-standard]
... bufferization and such...
```
And the rest of pipeline will stay unchanged.

With subgraphs they will have to extract existing `python-to-standard` stage, wrap both in subgraph and reinsert it into pipeline.

docs/rfcs/PipelineComposition.md
```diff
@@ -0,0 +1,170 @@
...
+
+## Proposal
+
+This API also allows a dynamic control flow between pipelines in graph, controlled by passes inside said pipelines.
+
+## New APIs
+
+### PipelineGraph
+```C++
+class PipelineGraph {
```
chencha3, line 30: It looks like to me not a graph, but a node which contains a sub-pipeline. It would be appreciated if you could make it clear in the naming.

PR #742, opened 2024-05-12
Add  plain layout support for XeGPU DPAS OP on VC mode (keep VNNI layout as internal representation) 

Jianhui-Li:
This PR adds plain layout representation for XeGPU DPAS OP for VC mode.  The VNNI layout is kept to as legacy representation. 

Motivation: 
1. Simplify XeGPU interface for the subgroup use case like XeTile. XeTile lowering now don't need to be responsible for the data layout consistency between VNNI and regular layout. There is no need for XeTile to differentiate A or B and apply VNNI axis and break the matrix form 2d to 3d.  
2. Simplify the support for VNNI layout consistent check. Exposing VNNI semantics to user means that it is user's responsibility to ensure whether the matrix has consistent VNNI layout for binary, matmul, or transpose/reduction op (should disallow).  
3. Unify the data representation for SIMT and SIMT model. The current 3d VNNI data representation in XeGPU level forces a mental model "VNNI transform" first then "WI data distribution".  The SIMT model actually is "WI data distribution" first and then "VNNI transform" on the data fragments.  By support 2d plain layout, user just need to deal with the "WI data distribution" on the SIMT lowering path, not involving 3d vector. 
For example, we first VNNI transform B<16x16xBF16> to B<8x16x2xBF16>, and then distribute to WI as <8x2xbf16>.   From SIMT point of view, it first distribute B<16x16xBF16> to <16x1xBF16>, then VNNI transform to <8x2xBF16>.  Introducing a VNNI layout <8x16x2xBF16> at the subgroup level and then distribute the 2nd dimension impacts the WI distribution rule with sg_map (which always assume 2d).  

Break changes: 
1. This PR simplifies the vnni attribute in the load_nd op. This is to unify SIMT and SIMD lowering path, since "VNNI=1" doesn't make any sense to SIMT path. Instead of setting the vnni_axis to be "1" or "0", user only need to set the "packed" attribute to be "true" or "false".  This brings breaking changes to XeGPU definition.  The load of matrix A will load to a 2d vector, and only reshape to 3d before DPAS assuming 3d vnni layout. 

2. This PR removes compile_hint ops. 

Additional Note:  
We keep the 3d VNNI data representation as legacy representation so this proposal doesn't minimize breaking changes to XeGPU OP definition.  The 2d plain layout is recommended to be the main usage for subgroup level XeGPU.  

The current 3d VNNI data representation doesn't keep the original layout, so it doesn't track the mapping to original matrix.  For example, the matrix A <8x32xbf16> may be represented as <8x16x2xbf16>, and add to a VNNI layout B matrix also represented as <8x16x2xbf16>.  The addition is illegal and should be detected as a compilation error. However, the current VNNI data representation can't be used to support a local analysis to determine such a error.  Adding a layout attribute to Vector dialect may solve the representation issue for VNNI layout, and we may replace the 3-d vector representation with the 2d vector with a VNNI attribute.


adam-smnk:
>  The load of matrix A will load to a 2d vector, and only reshape to 3d before DPAS assuming 3d vnni layout.

Is the reshape (or rather cast) needed at all?
DPAS could follow the design change and accept 2D shape for A operand. This would further reduce IR clutter and I don't think any info is lost.

Jianhui-Li:
> > The load of matrix A will load to a 2d vector, and only reshape to 3d before DPAS assuming 3d vnni layout.
> 
> Is the reshape (or rather cast) needed at all? DPAS could follow the design change and accept 2D shape for A operand. This would further reduce IR clutter and I don't think any info is lost.

I was thinking the same thing. I think we can allow DPAS to accept 2D for A and 3D for B.  @chencha3 do you have any concern on this?

Garra1980 (approved):

PR #767, opened 2024-06-07
Relax xegpu dpas operand requirement.

chencha3:
A and B operands of dpas could now be an either 2d or 3d vector.


Garra1980 (approved):
LGTM assuming CI is green

PR #773, opened 2024-06-10
VectorToXeGPU: Allows lowering vector.transfer_read and vector.transfer_write to XeGPU

Scarlet1ssimo:
- [x] Have you provided a meaningful PR description? Hope so.

---

This patch allows lowering `vector.transfer_read` and `vector.transfer_write`, which is quite common when handling the vectorization stuff, to corresponding XeGPU dialect.
Namely, it first create a descriptor then apply either a LoadNdOp or a StoreNdOp.
Directly accessing 1d vector runs into some unknown issues, specifically no error reported during compilation and run but got wrong memory access pattern. So temporarily, for access a 1d vector, it first translates to accessing a `1x?` vector, which then get reshaped back to 1d. At least it works as expected. Hope this could be fixed on some other sides.
Tested on PVC device.

### DIFF
--- a/test/Conversion/VectorToXeGPU/gemm_3x3.mlir
+++ b/test/Conversion/VectorToXeGPU/gemm_3x3.mlir
@@ -1,0 +1,114 @@
+// RUN: %python_executable %imex_runner --requires=l0-runtime -i %s --pass-pipeline-file=%p/vector-to-llvm.pp \
+// RUN:                                       --runner imex-cpu-runner -e main \
+// RUN:                                       --entry-point-result=void \
+// RUN:                                       --shared-libs=%mlir_runner_utils,%mlir_c_runner_utils,%levelzero_runtime --filecheck
+// RUN: %python_executable %imex_runner --requires=sycl-runtime -i %s --pass-pipeline-file=%p/vector-to-llvm.pp \
+// RUN:                                        --runner imex-cpu-runner -e main \
+// RUN:                                        --entry-point-result=void \
+// RUN:                                        --shared-libs=%mlir_runner_utils,%mlir_c_runner_utils,%sycl_runtime --filecheck
+#map = affine_map<(d0, d1) -> (0)>
+module @gemm attributes {gpu.container_module} {
+  memref.global "private" constant @__constant_3x3xi32_1 : memref<3x3xi32> = dense<1>
+  memref.global "private" constant @__constant_3x3xi32_0 : memref<3x3xi32> = dense<[[10, 11, 12], [13, 14, 15], [16, 17, 18]]>
+  memref.global "private" constant @__constant_3x3xi32 : memref<3x3xi32> = dense<[[1, 1, 1], [1, 1, 2], [3, 3, 3]]>
+  func.func @main() {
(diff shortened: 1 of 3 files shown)


silee2:
@Scarlet1ssimo I have a general question about the placement of this code.
Both the source and target dialect of this conversion pass are MLIR upstream dialect.
This conversion pass can be placed in upstream MLIR or as part of your target project.
Any specific reason IMEX is the best place for this pass?

lib/Conversion/VectorToXeGPU/VectorToXeGPU.cpp
```diff
@@ -0,0 +1,224 @@
...
+    auto loadType = VectorType::get(loadShape, resTileType.getElementType());
+    auto tDescTy =
+        xegpu::TensorDescType::get(loadShape, resTileType.getElementType());
+    mlir::SmallVector<mlir::OpFoldResult> tDescOffsets{read->getOperand(1),
+                                                       read->getOperand(2)};
+    rewriter.setInsertionPoint(read);
+    mlir::Value desc;
+    if (auto MemRefTypedSource =
+            mlir::cast<mlir::TypedValue<mlir::MemRefType>>(source)) {
+      desc = rewriter.create<mlir::xegpu::CreateNdDescOp>(
```
silee2, line 93: Shouldn't there be a check for memref rank here? XeGPU supports limited ranks.

lib/Conversion/VectorToXeGPU/VectorToXeGPU.cpp
```diff
@@ -0,0 +1,224 @@
...
+    rewriter.setInsertionPoint(write);
+    mlir::Value payload = write.getOperand(0);
+    if (rank == 1) {
+      payload = rewriter.create<vector::ShapeCastOp>(
+          write.getLoc(), intermediateType, write->getOperand(0));
+    }
+    mlir::Value desc;
+    if (auto MemRefTypedSource =
+            mlir::cast<mlir::TypedValue<mlir::MemRefType>>(source)) {
+      desc = rewriter.create<mlir::xegpu::CreateNdDescOp>(
```
silee2, line 175: Same as my comment above: Check rank and return failure if unsupported shape.

test/Conversion/VectorToXeGPU/gemm_3x3.mlir
```diff
@@ -0,0 +1,114 @@
+// RUN: %python_executable %imex_runner --requires=l0-runtime -i %s --pass-pipeline-file=%p/vector-to-llvm.pp \
```
silee2, line 1: This test in an Integration test running on GPU.
I would suggest placing in somewhere like under
test/Integration/Dialect/Vector/

test/Conversion/VectorToXeGPU/vector-to-llvm.pp
```diff
@@ -0,0 +1,18 @@
+builtin.module(
```
silee2, line 1: Should move to a folder for Integration tests along with the test case above.

silee2 (changes requested):

PR #803, opened 2024-07-10
ndarray lowering is compatible with one-shot-bufferize

tkarna:
- ndarray-to-linalg pass: all to_tensor ops use `restrict=true` to comply with one-shot-bufferize
- region-bufferize pass: to_tensor ops use `restrict=true`
- ndarray-to-linalg pass: fix reshape op copy handling


fschlimb (approved):
Yey!

silee2 (approved):
LGTM

PR #807, opened 2024-07-13
XeGPU RFC Update: chunk_size_per_lane as tensor_desc's attribute  

Jianhui-Li:
Associate chunk_size_per_lane with the tensor_desc, to facilitate the lowering of load_gather and store_scatter.  

Currently the lowering of load_gather/store_scatter infer the chunk_size_per_lane by computing number of elements being load by 16, which is error-prone.  "chunk_size_per_lane" should be included as part of tensor_desc, since it helps describes how the linear 1D memory is viewed as 2D memory (16x8xuint16 in the example below). 

  #tdesc_attr = !xegpu.tdesc_attr< memory_scope=slm, scattered=true, chunk_size_per_lane=8>
  %scatter_tdesc_chunk = XeGPU.create_tdesc, %base_addr, %offsets { mode = vc} :
		uint64, vector<16xindex> into tensor_desc<16x8xuint16, #tdesc_attr>

### DIFF
--- a/docs/rfcs/XeGPU.md
+++ b/docs/rfcs/XeGPU.md
@@ -206,9 +206,9 @@ Attributes `L1_hint`, `L2_hint`, `L3_hint`, and `memory_scope` can be applied to
 The example above creates a tensor_desc, which describes the memory base address and offsets for 16 uint8 values in the memory.  The number of work items (SIMD lanes) can be 1, 2, 4, 8, 16, 32.
 ```mlir
 
-  #tdesc_attr = !xegpu.tdesc_attr< memory_scope=slm, scattered=true>
+  #tdesc_attr = !xegpu.tdesc_attr< memory_scope=slm, scattered=true, chunk_size_per_lane=8>
   %scatter_tdesc_chunk = XeGPU.create_tdesc, %base_addr, %offsets
-		{chunk_size_per_lane=8, mode = vc} :
+		{mode = vc} :
 		uint64, vector<16xindex> into tensor_desc<16x8xuint16, #tdesc_attr>
 ```
 Attribute `memory_scope` indicates whether the tensor is located in the global (default) or shared local memory.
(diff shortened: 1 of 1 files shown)


Jianhui-Li:
@chencha3  @mshahneo @charithaintc  Could you please review this and approve it if you agree? It is a minor adjustment to facilitate the lowering.

docs/rfcs/XeGPU.md
```diff
@@ -222,11 +222,12 @@ Attribute `chunk_size_per_lane` specifies the size being loaded per each work it
         	  tensor_desc<16xuint8, #Scattered>, vector<16xi1> into vector<16xuint8>
 ```
 
-When loading a tensor_desc with chunk_size_per_lane attribute, the output vector must be 2D vector, with the chunk being treated as a new dimension. The consecutive 1D tensor data being loaded can be viewed as a 2D tensor loaded with transposition, with the chunk dimension transposed to the outer dimension.
+When a tensor_desc with chunk_size_per_lane attribute is loaded, the output vector must be 2D vector, with the chunk being treated as a new dimension. The consecutive 1D tensor data being loaded can be viewed as a 2D tensor loaded with transposition, with the chunk dimension transposed to the outer dimension. The transpose=[1,0] attribute must be present when chunk_size_per_lane is larger than 1.
```
charithaintc, line 225: why special case for chunk_size > 1. we can just say transpose=[1, 0] must always be present. right?

charithaintc (approved):
LGTM.

docs/rfcs/XeGPU.md
Jianhui-Li: if chunk_size = 1, then there is no transpose effect.  Or you mean the result should be always 2d as [chunk_size, inner_dim]. So if chunk_size =1, the result should be [1x16] instead of [16]?  What is the benefit of doing this?

charithaintc: yes. result is always 2d. the benefit is we don't need special verification code that needs to run for chunk_size == 1 case.

Jianhui-Li: OK. I just disallow the condition "chunk_size == 1" because that is useless.  User must not specify "chunk_size == 1".

chencha3 (approved):
LGTM.

Garra1980 (approved):

Garra1980:
@Jianhui-Li let's fix pre-commit check and merge

Jianhui-Li:
fixed now. @Garra1980

PR #877, opened 2024-09-14
[NFC] Adding tablegened dialect headers into IMEX_CONFIG_INCLUDE_DIRS

AllanZyne:
Added dialect headers into IMEX_CONFIG_INCLUDE_DIRS, so that other project can build their projects using imex build folder directly, without deploying it (save disk space)

### DIFF
--- a/cmake/modules/CMakeLists.txt
+++ b/cmake/modules/CMakeLists.txt
@@ -96,7 +96,7 @@ set(IMEX_CONFIG_INCLUDE_EXPORTS "include(\"\${IMEX_CMAKE_DIR}/IMEXTargets.cmake\
 set(IMEX_CONFIG_INCLUDE_DIRS
     "${IMEX_SOURCE_DIR}/include"
     "${IMEX_BINARY_DIR}/include"
-    "${IMEX_BINARY_DIR}/tools/Imex/include"
+    "${IMEX_GENERATED_HEADER_DIR}/include"
 )
 
 configure_file(
--- a/cmake/modules/CMakeLists.txt
+++ b/cmake/modules/CMakeLists.txt
@@ -96,6 +96,7 @@ set(IMEX_CONFIG_INCLUDE_EXPORTS "include(\"\${IMEX_CMAKE_DIR}/IMEXTargets.cmake\
 set(IMEX_CONFIG_INCLUDE_DIRS
     "${IMEX_SOURCE_DIR}/include"
     "${IMEX_BINARY_DIR}/include"
+    "${IMEX_BINARY_DIR}/tools/Imex/include"
 )
 
 configure_file(
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -265,7 +265,8 @@ endif()
 # IMEX headers in source tree
 include_directories(${CMAKE_CURRENT_SOURCE_DIR}/include)
 # Generated IMEX headers
-include_directories(${CMAKE_CURRENT_BINARY_DIR}/include)
+set(IMEX_GENERATED_HEADER_DIR ${CMAKE_CURRENT_BINARY_DIR})
+include_directories(${IMEX_GENERATED_HEADER_DIR}/include)
 message(STATUS "LLVM_DEFINITIONS: ${LLVM_DEFINITIONS}")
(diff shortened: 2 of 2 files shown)


AllanZyne:
Hi @fschlimb @tkarna, could you help to review this PR? 
Although this PR is not really related to sharpy functionality, but it's really helpful if it can be used it in developping sharpy.

fschlimb:
What does this do exactly? The dir does not exist.

fschlimb:
> ls tools/Imex/include/imex/Dialect/Dist/IR/

The path looks weird and it does not exist in my build:
```
(imex-dev) fschlimb@xxx:~/github/frameworks.ai.mlir.mlir-extensions$ ls build/tools/Imex/include/imex/Dialect/Dist/IR/
ls: cannot access 'build/tools/Imex/include/imex/Dialect/Dist/IR/': No such file or directory
(imex-dev) fschlimb@xxx:~/github/frameworks.ai.mlir.mlir-extensions$ ls build/tools/
CMakeFiles  cmake_install.cmake  imex-cpu-runner  imex-opt  imex-runner
```
But
```
(imex-dev) fschlimb@xxx:~/github/frameworks.ai.mlir.mlir-extensions$ ls build/include/imex/Dialect/Dist/IR/
CMakeFiles  cmake_install.cmake  DistOpsAttrs.cpp.inc  DistOpsAttrs.h.inc  DistOps.cpp.inc  DistOpsDialect.cpp.inc  DistOpsDialect.h.inc  DistOps.h.inc  DistOpsTypes.cpp.inc  DistOpsTypes.h.inc
```

cmake/modules/CMakeLists.txt
```diff
@@ -96,6 +96,7 @@ set(IMEX_CONFIG_INCLUDE_EXPORTS "include(\"\${IMEX_CMAKE_DIR}/IMEXTargets.cmake\
 set(IMEX_CONFIG_INCLUDE_DIRS
     "${IMEX_SOURCE_DIR}/include"
     "${IMEX_BINARY_DIR}/include"
+    "${IMEX_BINARY_DIR}/tools/Imex/include"
```
silee2, line 99: This path is only valid when building IMEX as an LLVM external project and setting Imex as the external project name.
I mean, using option 1 in README.md and passing something like
   -DLLVM_EXTERNAL_PROJECTS="Imex" \
   -DLLVM_EXTERNAL_IMEX_SOURCE_DIR=../mlir-extensions

silee2: You would need to change to level CMakeLists.txt
https://github.com/intel/mlir-extensions/blob/main/CMakeLists.txt#L268
To something like.
set(IMEX_GENERATED_HEADER_DIR ${CMAKE_CURRENT_BINARY_DIR})
include_directories(${IMEX_GENERATED_HEADER_DIR}/include)
And then change
https://github.com/intel/mlir-extensions/blob/main/cmake/modules/CMakeLists.txt#L98
to
"${IMEX_GENERATED_HEADER_DIR}/include"

silee2: @AllanZyne Thanks for catching this bug.

silee2:
> What does this do exactly? The dir does not exist.

@fschlimb It is a directory that is created when building IMEX as an LLVM external project. When you build LLVM and IMEX separately build tree structure is different, and 
 that directory is not created.
generated files from LLVM external projects are placed under
`${CMAKE_BINARY_DIR}/tools/<external project name>`
So generated headers for IMEX, if building as an LLVM external project will be located under
`${CMAKE_BINARY_DIR}/tools/Imex/include`
If you are building IMEX separately, generated headers will be in
`${CMAKE_BINARY_DIR}/include`
IMEX_BINARY_DIR == CMAKE_BINARY_DIR in both cases,
The current export logic works correctly for IMEX separate build, but not for IMEX as an LLVM external project. So it is actually a bug.

cmake/modules/CMakeLists.txt
AllanZyne: Thank you for your suggestions, @silee2. Done.

fschlimb (approved):

silee2 (approved):

PR #883, opened 2024-09-18
Support GEMM with f16 output.

antonio-cortes-perez:
With respect to the intrinsic that outputs f32, the change consists in encoding the result type (i.e., from f32 to f16). The full name of the intrinsic is: llvm.genx.dpas2.v128f16.v128i32.v64i32

Adding the parameter keepF16 to encodeVectorType allowed to keep the same behavior for the existing use cases.

Note: the enforceInteger parameter works only for F32 use cases. There are ops that use float encoding for f32 and int encoding for f16. For example, see below v128f32 vs v64i32:

xegpu.load_nd %4  : !xegpu.tensor_desc<8x16xf32> -> vector<8x16xf32>
generates
llvm.genx.raw.send2.v128f32.i1.v8i32

vs.

xegpu.load_nd %7 : !xegpu.tensor_desc<8x16xf16> -> vector<8x16xf16>
generates
llvm.genx.raw.send2.v64i32.i1.v8i32

### DIFF
--- a/include/imex/Utils/XeCommon.h
+++ b/include/imex/Utils/XeCommon.h
@@ -298,7 +298,7 @@ private:
 std::pair<std::string, mlir::VectorType>
 encodeVectorType(mlir::ConversionPatternRewriter &rewriter,
                  mlir::VectorType type, bool use64bitData = false,
-                 bool enforceInteger = false);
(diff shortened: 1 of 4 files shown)


Garra1980 (approved):
LGTM assuming CI is green

PR #887, opened 2024-09-20
add an option in insertGpuAllocs to skip the func args copy

zhczhong:
Add an option in insertGpuAllocs to skip the func args copy so that the user could decide whether a gpu::memcopy is needed for the input args


lib/Transforms/InsertGPUAllocs.cpp
```diff
@@ -545,7 +551,7 @@ class InsertGPUAllocsPass final
     // with gpu.alloc and insert a memref.copy from host to device
     for (const auto &it : gpuBufferParams) {
       auto param = block.getArgument(it.first);
-      if (isGpuAddrSpace(param))
+      if (isGpuAddrSpace(param) || isUsmArgs.getValue())
```
AndreyPavlenko, line 554: I think would be better to put the entire loop into `if (!isUsmArgs.getValue()) { for (...`.

AndreyPavlenko:
There is also pre-commit check failed.

lib/Transforms/InsertGPUAllocs.cpp
zhczhong: Thanks for the advice! I move the `if` out of the loop.(56cf8f1)

zhczhong:
> There is also pre-commit check failed.

The pre-commit check failure is not related to this PR but to other unverified commits. I fix them by [c7be8fc](https://github.com/intel/mlir-extensions/pull/887/commits/c7be8fc9b40ca5f720b23437cec0888dcd85257b)

AndreyPavlenko (approved):

PR #894, opened 2024-09-25
Update llvm commit

yifeizh2:
Update llvm commit to involve the fix for `tensor.unpack` canonicalization and `isaConvolutionOpInterface`.


yifeizh2:
@dchigarev The fixes mentioned in description are essential for the [layout propagation PR](https://github.com/intel/graph-compiler/pull/101) to pass CI check. The LLVM commit currently selected is chosen with the thought of making minimum modification (does not lead to any further changes to imex patch files, etc).
On the latest LLVM main branch, features in patch `0007-Move-chunk_size-into-TensorDesc.patch` has been merged. Not sure whether it will be better to directly update to the latest LLVM commit.

PR #916, opened 2024-10-09
add an option in insertGpuAllocs to skip the func args copy

zhczhong:
Add an option in insertGpuAllocs to skip the func args copy so that the user could decide whether a gpu::memcopy is needed for the input args

### DIFF
--- a/include/imex/Transforms/Passes.h
+++ b/include/imex/Transforms/Passes.h
@@ -19,12 +19,15 @@
 #include <memory>
 
 namespace imex {
+struct InsertGPUAllocsOptions;
 //===----------------------------------------------------------------------===//
 // Passes
 //===----------------------------------------------------------------------===//
 std::unique_ptr<mlir::Pass> createSerializeSPIRVPass();
 std::unique_ptr<mlir::Pass>
 createInsertGPUAllocsPass(const char *clientAPI = "vulkan");
+std::unique_ptr<mlir::Pass>
+createInsertGPUAllocsPass(const InsertGPUAllocsOptions &);
 std::unique_ptr<mlir::Pass> createSetSPIRVCapabilitiesPass();
 std::unique_ptr<mlir::Pass>
 createSetSPIRVAbiAttributePass(const char *clientAPI = "vulkan");
--- a/include/imex/Transforms/Passes.td
+++ b/include/imex/Transforms/Passes.td
@@ -41,7 +41,11 @@ def InsertGPUAllocs : Pass<"insert-gpu-allocs", "::mlir::func::FuncOp"> {
     Option<"clientAPI", "client-api", "std::string", /*default=*/"\"opencl\"",
            "The client API to use for inserting gpu allocs">,
     Option<"inRegions", "in-regions", "bool", "false",
-           "Add gpu allocs only for memref.AllocOps within GPU regions">
+           "Add gpu allocs only for memref.AllocOps within GPU regions">,
+        Option<"isUsmArgs", "is-usm-args", "bool", "false",
+           "Whether to use USM(unified shared memory) func args, in which the "
+           "host and device could access the same buffer and there is no need "
+           "to add memcpy explicitly">
   ];
 }
 
--- a/lib/Transforms/InsertGPUAllocs.cpp
+++ b/lib/Transforms/InsertGPUAllocs.cpp
@@ -47,6 +47,12 @@ public:
   explicit InsertGPUAllocsPass() : m_clientAPI("vulkan") {}
   explicit InsertGPUAllocsPass(const mlir::StringRef &clientAPI)
       : m_clientAPI(clientAPI) {}
+  explicit InsertGPUAllocsPass(const imex::InsertGPUAllocsOptions &options)
+      : InsertGPUAllocsBase<InsertGPUAllocsPass>(options) {
+    if (clientAPI == "opencl") {
+      m_clientAPI = "opencl";
+    }
+  }
 
   mlir::LogicalResult
   initializeOptions(mlir::StringRef options,
@@ -540,15 +546,17 @@ public:
     // This is the case where the inputs are passed as arguments to the
     // function. This code will add the IR for memory allocation on the device
     // with gpu.alloc and insert a memref.copy from host to device
-    for (const auto &it : gpuBufferParams) {
-      auto param = block.getArgument(it.first);
-      if (isGpuAddrSpace(param))
-        continue;
-      auto access = getAccessType(param);
-      access.hostRead = true;
-      access.hostWrite = true;
-      builder.setInsertionPointToStart(&block);
-      add_gpu_alloc(builder, param, access, term);
+    if (!isUsmArgs.getValue()) {
+      for (const auto &it : gpuBufferParams) {
+        auto param = block.getArgument(it.first);
+        if (isGpuAddrSpace(param))
+          continue;
+        auto access = getAccessType(param);
+        access.hostRead = true;
+        access.hostWrite = true;
+        builder.setInsertionPointToStart(&block);
+        add_gpu_alloc(builder, param, access, term);
+      }
     }
 
     // CallOp Case: This is the case where the memref producer is coming
@@ -579,5 +587,9 @@ private:
 namespace imex {
 std::unique_ptr<mlir::Pass> createInsertGPUAllocsPass(const char *clientAPI) {
   return std::make_unique<InsertGPUAllocsPass>(clientAPI);
+}
+std::unique_ptr<mlir::Pass>
+createInsertGPUAllocsPass(const InsertGPUAllocsOptions &option) {
+  return std::make_unique<InsertGPUAllocsPass>(option);
 }
 } // namespace imex
--- a/test/Transforms/BF16ToGPU/EltwiseAdd.bf16.mlir
+++ b/test/Transforms/BF16ToGPU/EltwiseAdd.bf16.mlir
@@ -65,3 +65,43 @@ module @eltwise_add attributes {gpu.container_module} {
   }
   func.func private @printMemrefBF16(memref<*xbf16>)
 }
+
+
+module @eltwise_add_usm attributes {gpu.container_module} {
+  memref.global "private" constant @__constant_10x20xbf16 : memref<10x20xbf16> = dense<5.000000e-01>
+  func.func @test(%arg0: memref<10x20xbf16>, %arg1: memref<10x20xbf16>) -> memref<10x20xbf16> {
+    %c20 = arith.constant 20 : index
+    %c10 = arith.constant 10 : index
+    %c1 = arith.constant 1 : index
+    %memref_1 = gpu.alloc  host_shared () : memref<10x20xbf16>
+    gpu.launch_func  @test_kernel::@test_kernel blocks in (%c10, %c20, %c1) threads in (%c1, %c1, %c1)  args(%arg0 : memref<10x20xbf16>, %arg1 : memref<10x20xbf16>, %memref_1 : memref<10x20xbf16>)
+    %alloc = memref.alloc() : memref<10x20xbf16>
+    memref.copy %memref_1, %alloc : memref<10x20xbf16> to memref<10x20xbf16>
+    gpu.dealloc  %memref_1 : memref<10x20xbf16>
+    return %alloc : memref<10x20xbf16>
+  }
+  gpu.module @test_kernel attributes {spirv.target_env = #spirv.target_env<#spirv.vce<v1.0, [Addresses, Float16Buffer, Int64, Int16, Int8, Kernel, Linkage, Vector16, GenericPointer, Groups, Float16, Float64, AtomicFloat32AddEXT, ExpectAssumeKHR, SubgroupDispatch, VectorComputeINTEL, VectorAnyINTEL, Bfloat16ConversionINTEL], [SPV_EXT_shader_atomic_float_add, SPV_KHR_expect_assume, SPV_INTEL_vector_compute, SPV_INTEL_bfloat16_conversion]>, api=OpenCL, #spirv.resource_limits<>>} {
+    gpu.func @test_kernel(%arg0: memref<10x20xbf16>, %arg1: memref<10x20xbf16>, %arg2: memref<10x20xbf16>) kernel attributes {VectorComputeFunctionINTEL, gpu.known_block_size = array<i32: 1, 1, 1>, gpu.known_grid_size = array<i32: 10, 20, 1>, spirv.entry_point_abi = #spirv.entry_point_abi<>} {
+      %block_id_x = gpu.block_id  x
+      %block_id_y = gpu.block_id  y
+      %cst = arith.constant 0.5 : bf16
+      %0 = memref.load %arg0[%block_id_x, %block_id_y] : memref<10x20xbf16>
+      %1 = memref.load %arg1[%block_id_x, %block_id_y] : memref<10x20xbf16>
+      %2 = arith.addf %0, %1 : bf16
+      %3 = arith.addf %2, %cst : bf16
+      memref.store %3, %arg2[%block_id_x, %block_id_y] : memref<10x20xbf16>
+      gpu.return
+    }
+  }
+  func.func @main() {
+    %0 = memref.get_global @__constant_10x20xbf16 : memref<10x20xbf16>
+    %1 = memref.get_global @__constant_10x20xbf16 : memref<10x20xbf16>
+    %2 = call @test(%0, %1) : (memref<10x20xbf16>, memref<10x20xbf16>) -> memref<10x20xbf16>
+    %cast = memref.cast %2 : memref<10x20xbf16> to memref<*xbf16>
+    // CHECK: Unranked Memref base@ = {{(0x)?[-9a-f]*}}
+    // CHECK-COUNT-200: 1.5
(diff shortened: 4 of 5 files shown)


silee2:
I don't think this is the right solution.
If caller functions args are directly used,
Gpu mem allocation and copy is only need if the arg's (a memref) memspace is not gpu.
USM vs Device mem shouldn't matter in this case.
Better way would be to use 
https://mlir.llvm.org/docs/Dialects/GPU/#gpu-address-spaces
For marking gpu or usm memory args.
And then in this pass skip func arg alloc and copy based on the address space attribute.

zhczhong:
> I don't think this is the right solution. If caller functions args are directly used, Gpu mem allocation and copy is only need if the arg's (a memref) memspace is not gpu. USM vs Device mem shouldn't matter in this case. Better way would be to use https://mlir.llvm.org/docs/Dialects/GPU/#gpu-address-spaces For marking gpu or usm memory args. And then in this pass skip func arg alloc and copy based on the address space attribute.

If the memory is usm-shared and we mark it with gpu-address-spaces, the copy on the device side is indeed skipped but the memory access on the host side will still need an additional explicit copy.

kurapov-peter:
> I don't think this is the right solution. If caller functions args are directly used, Gpu mem allocation and copy is only need if the arg's (a memref) memspace is not gpu. USM vs Device mem shouldn't matter in this case. Better way would be to use https://mlir.llvm.org/docs/Dialects/GPU/#gpu-address-spaces For marking gpu or usm memory args. And then in this pass skip func arg alloc and copy based on the address space attribute.

Marking host-side usm memory with a gpu address space is semantically incorrect. You are relying on the cpu lowering to ignore a gpu address space that shouldn't have been there in the first place. The solution to mark the callee only (this is what actually happens after the lowering) would be feasible if not for type-checking between the kernel caller and the callee. This is a simple way to avoid the problem and does not change the default behavior.

silee2:
> > I don't think this is the right solution. If caller functions args are directly used, Gpu mem allocation and copy is only need if the arg's (a memref) memspace is not gpu. USM vs Device mem shouldn't matter in this case. Better way would be to use https://mlir.llvm.org/docs/Dialects/GPU/#gpu-address-spaces For marking gpu or usm memory args. And then in this pass skip func arg alloc and copy based on the address space attribute.
> 
> Marking host-side usm memory with a gpu address space is semantically incorrect. You are relying on the cpu lowering to ignore a gpu address space that shouldn't have been there in the first place. The solution to mark the callee only (this is what actually happens after the lowering) would be feasible if not for type-checking between the kernel caller and the callee. This is a simple way to avoid the problem and does not change the default behavior.

Makes sense.
Unfortunately, MLIR does not have an address space attribute for USM memory.
Pass option is not fail safe either. In assumes, all caller direct forwarded args are USM memory. There is no way for the pass to check, so user needs to make that guarantee. Especially if there are multiple callees and callers involved. I will approve but would like to see one more approval before merging.

silee2 (approved):

silee2 (changes requested):
Found one thing I overlooked.
Please add a test case for the new option.

kurapov-peter:
> > > I don't think this is the right solution. If caller functions args are directly used, Gpu mem allocation and copy is only need if the arg's (a memref) memspace is not gpu. USM vs Device mem shouldn't matter in this case. Better way would be to use https://mlir.llvm.org/docs/Dialects/GPU/#gpu-address-spaces For marking gpu or usm memory args. And then in this pass skip func arg alloc and copy based on the address space attribute.
> 
> > 
> 
> > Marking host-side usm memory with a gpu address space is semantically incorrect. You are relying on the cpu lowering to ignore a gpu address space that shouldn't have been there in the first place. The solution to mark the callee only (this is what actually happens after the lowering) would be feasible if not for type-checking between the kernel caller and the callee. This is a simple way to avoid the problem and does not change the default behavior.
> 
> 
> 
> Makes sense.
> 
> Unfortunately, MLIR does not have an address space attribute for USM memory.
> 
> Pass option is not fail safe either. In assumes, all caller direct forwarded args are USM memory. There is no way for the pass to check, so user needs to make that guarantee. Especially if there are multiple callees and callers involved. I will approve but would like to see one more approval before merging.
> 
> 

Actually, if we care about guarantees, there's an option to insert runtime checks. For the case at hand we know for sure those are usm allocations, so this is just overhead.

As to usm address space, this could be a viable solution on the long run. However, we'd need to revisit llvm type converter behavior as it makes assumptions about the default address space.

zhczhong:
> Found one thing I overlooked. Please add a test case for the new option "is-usm-args"

Thanks for reminding! A test case is added

silee2 (approved):

mshahneo:
Hi @zhczhong ,

An important addition, thank you so much :).
Could you please add a bf16 end-to-end test case? It could be a very simple test case, but uses the bf16-to-gpu pass. Wanted to check how it would interact with bf16-to-gpu transformation.
A reference to a simple test case is: https://github.com/intel/mlir-extensions/blob/main/test/Integration/Dialect/Gpu/EltwiseAdd_BF16.mlir.

You could just create a similar USM version of this test or you can create something else :)

mshahneo:
> > > I don't think this is the right solution. If caller functions args are directly used, Gpu mem allocation and copy is only need if the arg's (a memref) memspace is not gpu. USM vs Device mem shouldn't matter in this case. Better way would be to use https://mlir.llvm.org/docs/Dialects/GPU/#gpu-address-spaces For marking gpu or usm memory args. And then in this pass skip func arg alloc and copy based on the address space attribute.
> > 
> > 
> > Marking host-side usm memory with a gpu address space is semantically incorrect. You are relying on the cpu lowering to ignore a gpu address space that shouldn't have been there in the first place. The solution to mark the callee only (this is what actually happens after the lowering) would be feasible if not for type-checking between the kernel caller and the callee. This is a simple way to avoid the problem and does not change the default behavior.
> 
> Makes sense. Unfortunately, MLIR does not have an address space attribute for USM memory. Pass option is not fail safe either. In assumes, all caller direct forwarded args are USM memory. There is no way for the pass to check, so user needs to make that guarantee. Especially if there are multiple callees and callers involved. I will approve but would like to see one more approval before merging.

This may be out of scope of this PR.
But the concern raised by SangIk is a valid one, adding this option to the pass does give us the option to not create gpu-allocs and memcpy for the USM memory, but adds a very tricky pitfall. If the user uses this option, then they need to absolutely make sure that, all their host function uses USM arguments that gets passed to the device. Otherwise, it would result in runtime crash/segfault.  

What do you think if we could add an attribute to XeGPU (similar to XeGPU_MemoryScope) for USM, and mark memrefs with this attribute whenever we pass a USM memref to a function. insert-gpu-alloc could inturn use that attribute.

zhczhong:
> Hi @zhczhong ,
> 
> An important addition, thank you so much :). Could you please add a bf16 end-to-end test case? It could be a very simple test case, but uses the bf16-to-gpu pass. Wanted to check how it would interact with bf16-to-gpu transformation. A reference to a simple test case is: https://github.com/intel/mlir-extensions/blob/main/test/Integration/Dialect/Gpu/EltwiseAdd_BF16.mlir.
> 
> You could just create a similar USM version of this test or you can create something else :)

Thanks for the suggestion! The test has been added but the `bf16-to-gpu` indeed failed to process the USM case.

```mlir
BF16ToGPU/EltwiseAdd.bf16.mlir:81:5: error: 'func.return' op has 1 operands, but enclosing function (@test) returns 0
    return %alloc : memref<10x20xbf16>
    ^
"builtin.module"() <{sym_name = "eltwise_add_usm"}> ({
    "memref.global"() <{constant, initial_value = dense<5.000000e-01> : tensor<10x20xbf16>, sym_name = "__constant_10x20xbf16", sym_visibility = "private", type = memref<10x20xbf16>}> : () -> ()
    "func.func"() <{function_type = (memref<10x20xi16>, memref<10x20xi16>) -> (), sym_name = "test"}> ({
    ^bb0(%arg3: memref<10x20xi16>, %arg4: memref<10x20xi16>):
      %23 = "arith.constant"() <{value = 20 : index}> : () -> index
      %24 = "arith.constant"() <{value = 10 : index}> : () -> index
      %25 = "arith.constant"() <{value = 1 : index}> : () -> index
      %26 = "arith.constant"() <{value = 0 : index}> : () -> index
      %27 = "gpu.alloc"() <{hostShared, operandSegmentSizes = array<i32: 0, 0, 0>}> : () -> memref<400xi8>
      %28 = "memref.view"(%27, %26) : (memref<400xi8>, index) -> memref<10x20xbf16>
      %29 = "memref.view"(%27, %26) : (memref<400xi8>, index) -> memref<10x20xi16>
      "gpu.launch_func"(%24, %23, %25, %25, %25, %25, %arg3, %arg4, %29) <{kernel = @test_kernel::@test_kernel, operandSegmentSizes = array<i32: 0, 1, 1, 1, 1, 1, 1, 0, 0, 0, 0, 3, 0>}> : (index, index, index, index, index, index, memref<10x20xi16>, memref<10x20xi16>, memref<10x20xi16>) -> ()
      %30 = "memref.alloc"() <{operandSegmentSizes = array<i32: 0, 0>}> : () -> memref<10x20xbf16>
      "memref.copy"(%28, %30) : (memref<10x20xbf16>, memref<10x20xbf16>) -> ()
      "gpu.dealloc"(%27) : (memref<400xi8>) -> ()
      "func.return"(%30) : (memref<10x20xbf16>) -> ()
    }) : () -> ()
    "gpu.module"() <{sym_name = "test_kernel"}> ({
      "gpu.func"() <{function_type = (memref<10x20xi16>, memref<10x20xi16>, memref<10x20xi16>) -> ()}> ({
      ^bb0(%arg0: memref<10x20xi16>, %arg1: memref<10x20xi16>, %arg2: memref<10x20xi16>):
        %4 = "gpu.block_id"() <{dimension = #gpu<dim x>}> : () -> index
        %5 = "gpu.block_id"() <{dimension = #gpu<dim y>}> : () -> index
        %6 = "arith.constant"() <{value = 16128 : i16}> : () -> i16
        %7 = "memref.load"(%arg0, %4, %5) : (memref<10x20xi16>, index, index) -> i16
        %8 = "memref.load"(%arg1, %4, %5) : (memref<10x20xi16>, index, index) -> i16
        %9 = "arith.bitcast"(%7) : (i16) -> bf16
        %10 = "arith.extf"(%9) : (bf16) -> f32
        %11 = "arith.bitcast"(%8) : (i16) -> bf16
        %12 = "arith.extf"(%11) : (bf16) -> f32
        %13 = "arith.addf"(%10, %12) <{fastmath = #arith.fastmath<none>}> : (f32, f32) -> f32
        %14 = "arith.truncf"(%13) : (f32) -> bf16
        %15 = "arith.bitcast"(%14) : (bf16) -> i16
        %16 = "arith.bitcast"(%15) : (i16) -> bf16
        %17 = "arith.extf"(%16) : (bf16) -> f32
        %18 = "arith.bitcast"(%6) : (i16) -> bf16
        %19 = "arith.extf"(%18) : (bf16) -> f32
        %20 = "arith.addf"(%17, %19) <{fastmath = #arith.fastmath<none>}> : (f32, f32) -> f32
        %21 = "arith.truncf"(%20) : (f32) -> bf16
        %22 = "arith.bitcast"(%21) : (bf16) -> i16
        "memref.store"(%22, %arg2, %4, %5) : (i16, memref<10x20xi16>, index, index) -> ()
        "gpu.return"() : () -> ()
      }) {VectorComputeFunctionINTEL, gpu.kernel, gpu.known_block_size = array<i32: 1, 1, 1>, gpu.known_grid_size = array<i32: 10, 20, 1>, spirv.entry_point_abi = #spirv.entry_point_abi<>, sym_name = "test_kernel", workgroup_attributions = 0 : i64} : () -> ()
    }) {spirv.target_env = #spirv.target_env<#spirv.vce<v1.0, [Addresses, Float16Buffer, Int64, Int16, Int8, Kernel, Linkage, Vector16, GenericPointer, Groups, Float16, Float64, AtomicFloat32AddEXT, ExpectAssumeKHR, SubgroupDispatch, VectorComputeINTEL, VectorAnyINTEL, Bfloat16ConversionINTEL], [SPV_EXT_shader_atomic_float_add, SPV_KHR_expect_assume, SPV_INTEL_vector_compute, SPV_INTEL_bfloat16_conversion]>, api=OpenCL, #spirv.resource_limits<>>} : () -> ()
    "func.func"() <{function_type = () -> (), sym_name = "main"}> ({
      %0 = "memref.get_global"() <{name = @__constant_10x20xbf16}> : () -> memref<10x20xbf16>
      %1 = "memref.get_global"() <{name = @__constant_10x20xbf16}> : () -> memref<10x20xbf16>
      %2 = "func.call"(%0, %1) <{callee = @test}> : (memref<10x20xbf16>, memref<10x20xbf16>) -> memref<10x20xbf16>
      %3 = "memref.cast"(%2) : (memref<10x20xbf16>) -> memref<*xbf16>
      "func.call"(%3) <{callee = @printMemrefBF16}> : (memref<*xbf16>) -> ()
      "func.return"() : () -> ()
    }) : () -> ()
    "func.func"() <{function_type = (memref<*xbf16>) -> (), sym_name = "printMemrefBF16", sym_visibility = "private"}> ({
    }) {llvm.emit_c_interface} : () -> ()
  }) {gpu.container_module} : () -> ()
```

zhczhong:
> > > > I don't think this is the right solution. If caller functions args are directly used, Gpu mem allocation and copy is only need if the arg's (a memref) memspace is not gpu. USM vs Device mem shouldn't matter in this case. Better way would be to use https://mlir.llvm.org/docs/Dialects/GPU/#gpu-address-spaces For marking gpu or usm memory args. And then in this pass skip func arg alloc and copy based on the address space attribute.
> > > 
> > > 
> > > Marking host-side usm memory with a gpu address space is semantically incorrect. You are relying on the cpu lowering to ignore a gpu address space that shouldn't have been there in the first place. The solution to mark the callee only (this is what actually happens after the lowering) would be feasible if not for type-checking between the kernel caller and the callee. This is a simple way to avoid the problem and does not change the default behavior.
> > 
> > 
> > Makes sense. Unfortunately, MLIR does not have an address space attribute for USM memory. Pass option is not fail safe either. In assumes, all caller direct forwarded args are USM memory. There is no way for the pass to check, so user needs to make that guarantee. Especially if there are multiple callees and callers involved. I will approve but would like to see one more approval before merging.
> 
> This may be out of scope of this PR. But the concern raised by SangIk is a valid one, adding this option to the pass does give us the option to not create gpu-allocs and memcpy for the USM memory, but adds a very tricky pitfall. If the user uses this option, then they need to absolutely make sure that, all their host function uses USM arguments that gets passed to the device. Otherwise, it would result in runtime crash/segfault.
> 
> What do you think if we could add an attribute to XeGPU (similar to XeGPU_MemoryScope) for USM, and mark memrefs with this attribute whenever we pass a USM memref to a function. insert-gpu-alloc could inturn use that attribute.

Yes, marking these memref with a usm memory space could keep the semantic correct and let the pass behavior more precise. But it also require us to @kurapov-peter What do you think about this way?

zhczhong:
> > > > I don't think this is the right solution. If caller functions args are directly used, Gpu mem allocation and copy is only need if the arg's (a memref) memspace is not gpu. USM vs Device mem shouldn't matter in this case. Better way would be to use https://mlir.llvm.org/docs/Dialects/GPU/#gpu-address-spaces For marking gpu or usm memory args. And then in this pass skip func arg alloc and copy based on the address space attribute.
> > > 
> > > 
> > > Marking host-side usm memory with a gpu address space is semantically incorrect. You are relying on the cpu lowering to ignore a gpu address space that shouldn't have been there in the first place. The solution to mark the callee only (this is what actually happens after the lowering) would be feasible if not for type-checking between the kernel caller and the callee. This is a simple way to avoid the problem and does not change the default behavior.
> > 
> > 
> > Makes sense. Unfortunately, MLIR does not have an address space attribute for USM memory. Pass option is not fail safe either. In assumes, all caller direct forwarded args are USM memory. There is no way for the pass to check, so user needs to make that guarantee. Especially if there are multiple callees and callers involved. I will approve but would like to see one more approval before merging.
> 
> This may be out of scope of this PR. But the concern raised by SangIk is a valid one, adding this option to the pass does give us the option to not create gpu-allocs and memcpy for the USM memory, but adds a very tricky pitfall. If the user uses this option, then they need to absolutely make sure that, all their host function uses USM arguments that gets passed to the device. Otherwise, it would result in runtime crash/segfault.
> 
> What do you think if we could add an attribute to XeGPU (similar to XeGPU_MemoryScope) for USM, and mark memrefs with this attribute whenever we pass a USM memref to a function. insert-gpu-alloc could inturn use that attribute.

Yes, marking the memref with a usm memory space could make the semantic correct and let the pass behavior more precise. But it also requires the front end  doing the corresponding change. @kurapov-peter What do you think abut this way?

kurapov-peter:
> What do you think if we could add an attribute to XeGPU (similar to XeGPU_MemoryScope) for USM, and mark memrefs with this attribute whenever we pass a USM memref to a function. insert-gpu-alloc could inturn use that attribute.

I agree there should be a proper attribute that would ensure type consistency for USM buffers (and not violate host-side semantic). XeGPU does not seem a good place for it since that would be the same as putting a gpu attribute to a host-side buffer. Host-device shared memory is not unique to intel gpus, the abstraction should be vendor-agnostic.

PR #939, opened 2024-10-23
Add compile.sh to CI

leshikus:
Subj


leshikus:
@Garra1980  we need to keep compile.sh as is, and change the build dir name instead to match another repo

Garra1980:
> @Garra1980 we need to keep `compile.sh` as is, and change the `_build` dir name instead to match another repo

I agree, trying to see whether build is okay

.github/workflows/build.yml
```diff
@@ -37,51 +36,40 @@ jobs:
...
-          cat $LLVM_SHA_FILE >> $GITHUB_ENV
-          echo 'EOF' >> $GITHUB_ENV
+          echo LLVM_SHA=$(cat build_tools/llvm_version.txt) | tee -a $GITHUB_ENV
 
       - name: Cache LLLVM-MLIR
         id: cache-llvm-mlir
         uses: actions/cache@v4
         env:
-          LLVM_CACHE_NUMBER: 1  # Increase to reset cache
+          LLVM_CACHE_NUMBER: 2  # Increase to reset cache
```
Garra1980, line 57: do we really need this one?

scripts/compile.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/sh
+
+set -e
+set -vx
+
+cd $(dirname "$0")/..
+
+cmake -S . -B build -GNinja  -DCMAKE_BUILD_TYPE=Release -DCMAKE_C_COMPILER=/usr/bin/clang -DCMAKE_CXX_COMPILER=/usr/bin/clang++ -DLLVM_ENABLE_ASSERTIONS=ON -DLLVM_USE_LINKER=gold -DLLVM_ENABLE_ZSTD=OFF "$@"
```
Garra1980, line 8: let's use gcc not clang as we do it currently

Garra1980 (approved):

PR #940, opened 2024-10-23
LSC 2D prefetch: Unify type of dummy data arg to i32

silee2:
since the generated intinrsic call is independent of dummy data arg type. Type mismatch error happens if multiple callers use different types for dummy data arg.
For non 32bit element type, normalize 2D prefetch shape around i32 data type. Inner dimension gets scaled according to ratio between 32 and bitwidth of element type
Add mixed type 2D prefetch test case.


mshahneo (approved):
LGTM

Garra1980 (approved):

PR #973, opened 2024-12-03
[Region] Bufferize with one-shot-bufferize

tkarna:
- Remove region-bufferize pass and Region dialect conversion patterns
- Add BufferizableOpInterface extensions to Region dialect to enable bufferization via one-shot-bufferize
- Update region bufferize test that uses one-shot-bufferize

---


silee2 (approved):

lib/Dialect/Region/Transforms/BufferizableOpInterfaceImpl.cpp
```diff
@@ -0,0 +1,178 @@
...
+  }
+
+  ::mlir::LogicalResult
+  bufferize(::mlir::Operation *op, ::mlir::RewriterBase &rewriter,
+            const ::mlir::bufferization::BufferizationOptions &options) const {
+    auto envOp = ::mlir::cast<region::EnvironmentRegionOp>(op);
+    if (noTensorsIn(envOp.getArgs()) && noTensorsIn(envOp.getResults())) {
+      // Nothing to do.
+      return ::mlir::success();
+    }
```
fschlimb, line 87: Is this needed? the docs say
```
         This method will never be called on ops that do not have at least one
         tensor operand/result.
```

lib/Dialect/Region/Transforms/BufferizableOpInterfaceImpl.cpp
```diff
@@ -0,0 +1,178 @@
...
+    return ::llvm::isa<::mlir::TensorType>(v.getType());
+  };
+  if (::llvm::any_of(values, isTensor))
+    return false;
+  return true;
+}
+
+/// Convert values to buffers. If a value is a tensor, get a buffer for it.
+::mlir::LogicalResult
+convertToBuffers(::mlir::ValueRange values,
```
fschlimb, line 30: Why do you need this? Shouldn't `getBuffer` be sufficient?

lib/Dialect/Region/Transforms/BufferizableOpInterfaceImpl.cpp
```diff
@@ -0,0 +1,178 @@
...
+
+/// Bufferization of region.env_region_yield. Replaced with a new yield that
+/// operates on memrefs.
+struct EnvironmentRegionYieldOpInterface
+    : public ::mlir::bufferization::BufferizableOpInterface::ExternalModel<
+          EnvironmentRegionYieldOpInterface, region::EnvironmentRegionYieldOp> {
+  bool bufferizesToMemoryRead(
+      ::mlir::Operation *op, ::mlir::OpOperand &opOperand,
+      const ::mlir::bufferization::AnalysisState &state) const {
+    return true;
```
fschlimb, line 126: what is it reading?

tkarna: This implementation follows ` scf.yield` op bufferization: read is true, write is false etc.

lib/Dialect/Region/Transforms/BufferizableOpInterfaceImpl.cpp
tkarna: Agreed, this is not needed.

fschlimb (approved):

PR #1021, opened 2025-02-09
Update sg_map distribution rules for XeGPU.md 

Jianhui-Li:
The previous sg_map distribution rule assumes it occurs after the VNNI propagation pass, meaning the input tensor could be 3d with the VNNI layout.

The new distribution rule applies focus on 2d tensors, where each work item (WI) receives a 2d tensor. The outer dimension is lowered to the number of SIMT vectors, and the inner dimension is lowered to the number of values that need to be packed.
@charithaintc  @akroviakov 

### DIFF
--- a/docs/rfcs/XeGPU.md
+++ b/docs/rfcs/XeGPU.md
@@ -534,7 +534,7 @@ For matrix load with transpose for A or B
 A simple rule of thumb is that wi_data size is 16 bit for matrix a (with exception for tf32 data type) on PVC. For all rest mapping, the wi_data size is 32bit, regardless PVC or ARC.
 Reference of related spirv extension: [SPV_INTEL_2d_block_io](https://github.com/KhronosGroup/SPIRV-Registry/pull/305), [add SPV_INTEL_subgroup_matrix_multiply_accumulate](https://github.com/KhronosGroup/SPIRV-Registry/pull/306)
 
-The sg_map required by DPAS operation can be propogated from DPAS to 2D block load operations. The sg_map can be associated with sg_map temporarily and throw away after the sg_map is propogated to tensor descriptor. The sg_map does not describe the data fragments  after the tensor being loaded to register, which is already being packed. Instead, it describes the data fragments based on the unpacked plain layout of the input and output tensors.  
+The sg_map required by DPAS operation can be propogated from DPAS to 2D block load operations at subgroup level. The sg_map can be associated with sg_map temporarily and throw away after the sg_map is propogated to tensor descriptor. The sg_map does not describe the data fragments  after the tensor being loaded to register, which is already being packed. Instead, it describes the data fragments based on the unpacked plain layout of the input and output tensors.  
 
 Below are a few examples of DPAS's sg_map.
 ```mlir
@@ -543,6 +543,7 @@ Below are a few examples of DPAS's sg_map.
   #sg_map_c = xegpu.sg_map<wi_layout = [1, 16], wi_data = [1, 1]>
   #sg_map_b = xegpu.sg_map<wi_layout = [1, 16], wi_data = [2, 1]>
 
+  // Before WI distribution: %vector_c = xegpu.dpas %vector_a, %vector_b {#sg_map_a #sg_map_b_reg #sg_map_c} :vector<8x16xbf16>, vector<16x16xbf16> into vector<8x16xfloat>
   %vector_c = xegpu.dpas %vector_a, %vector_b {#sg_map_a #sg_map_b_reg #sg_map_c} :vector<8x1xbf16>, vector<8x2xbf16> into vector<8x1xfloat>
 
   ARC int8 example
@@ -550,20 +551,64 @@ Below are a few examples of DPAS's sg_map.
   #sg_map_c = xegpu.sg_map<wi_layout = [1, 8], wi_data = [1, 1]>
   #sg_map_b_ui8_reg = xegpu.sg_map<wi_layout = [1, 8], wi_data = [4, 1]>
 
-  %vector_c = xegpu.dpas %vector_a, %vector_b {#sg_map_a_ui #sg_map_b_ui8_reg #sg_map_c} :vector<8x4xui8>, vector<8x4xui8> into vector<8x1xfloat>
+  // Before WI distribution: %vector_c = xegpu.dpas %vector_a, %vector_b {#sg_map_a #sg_map_b_reg #sg_map_c} :vector<8x32xui8>, vector<32x16xui8> into vector<8x16xi32>
+  %vector_c = xegpu.dpas %vector_a, %vector_b {#sg_map_a_ui #sg_map_b_ui8_reg #sg_map_c} :vector<8x4xui8>, vector<8x4xui8> into vector<8x1xi32>
 ```
 
-The sg_map propagation process may encounter other operations that require modifications to the mapping between the input and output operands. Specifically, the transpose operation swaps the wi_layout and wi_data to correctly track the data fragments affected by the transpose. The bitcast operation adjusts the wi_data to reflect the correct number of data elements being packed. Operations such as reduction, broadcast, and regular element-wise operations do not modify the sg_map.
+The sg_map propagation process may encounter other operations that require modifications to the mapping between the input and output operands. Specifically, the transpose operation swaps the wi_layout and wi_data to correctly track the data fragments affected by the transpose. The bitcast operation adjusts the wi_data to reflect the correct number of data elements being packed. 
 
 ```mlir
   #sg_map_a = xegpu.sg_map<wi_layout = [1, 16], wi_data = [1, 2]> //the producer op of vector_a uses #sg_map_a
   #sg_map_b = xegpu.sg_map<wi_layout = [1, 16], wi_data = [2, 1]>
-  //Before WI distribution:  %vector_b = vector.transpose %vector_a {#sg_map_b} :vector<16x16xbf16> into vector<16x16xbf16> 
+  //Before WI distribution:  %vector_b = vector.transpose [1, 0] %vector_a {#sg_map_b} :vector<16x16xbf16> into vector<16x16xbf16> 
   %vector_b = vector.transpose %vector_a {#sg_map_b} :vector<8x2xbf16> into vector<8x2xbf16>
+
+  #sg_map_a = xegpu.sg_map<wi_layout = [1, 16], wi_data = [1, 1]> //the producer op of vector_a uses #sg_map_a
+  #sg_map_b = xegpu.sg_map<wi_layout = [1, 16], wi_data = [1, 2]>
+  //Before WI distribution:  %vector_b = vector.bitcast %vector_a {#sg_map_b} :vector<8x16xi16> into vector<8x32xi8> 
+  %vector_b = vector.bitcast %vector_a :vector<8x1xi16> into vector<8x2xi8>
 ```
 
+Operations such as reduction and broadcast are exception to WI distribution rule. The vector of the reduced dimension doesn't participate in the WI distribution so that the dimension size doesn't change. The wi_layout must be [1, %subgroup_size], and wi_data must be [1, 1]. The SIMT distribution needs to use different dialect operations since the vector dialect can't express cross-lane semantics.  
+```mlir
 
-user must use for the WI data distribution of 1D block load and regular load with chunk_size on PVC and ARC. Not using this sg_map defined here leads to undefined behavior.
+  #sg_map_a = xegpu.sg_map<wi_layout = [1, 16], wi_data = [1, 1]> //the producer op of vector_a uses #sg_map_a
+  #sg_map_b = xegpu.sg_map<wi_layout = [1, 16], wi_data = [1, 1]>
+  //Before WI distribution:  %vector_b = xetile.reduction [1] %vector_a {#sg_map_b} :vector<16x16xbf16> into vector<16x1xbf16> //xetile.reduction used as vector.reduction doesn't keep the reduction dim in the result vector
+  %vector_b = gpu.subgroup_reduce %vector_a :vector<16x1xbf16> into vector<16x1xbf16>
+
+  #sg_map_a = xegpu.sg_map<wi_layout = [1, 16], wi_data = [1, 1]> //the producer op of vector_a uses #sg_map_a
+  #sg_map_b = xegpu.sg_map<wi_layout = [1, 16], wi_data = [1, 1]>
+  //Before WI distribution:  %vector_b = xetile.broadcast %vector_a {#sg_map_b} :vector<16x1xbf16> into vector<16x16xbf16> //xetile.broadcast used as vector.broadcast doesn't allow the broadcast dim in the input vector
+  %cst0 = arith.constant 0 : i32
+  %7, %8 = gpu.shuffle idx %0, %cst0, %subgroup_size: vector<16x1xbf16> into vector<16x1xbf16>
+
+```
+
+Extract, insert and regular element-wise operations do not modify the sg_map. However, the result of SIMT distribution may have to use different operations. 
+```mlir
+  #sg_map_a = xegpu.sg_map<wi_layout = [1, 16], wi_data = [2, 1]> //the producer op of vector_a uses #sg_map_a
+  #sg_map_b = xegpu.sg_map<wi_layout = [1, 16], wi_data = [2, 1]>
+  // Before WI distribution:  %vector_b = vector.extract %vector_a [0] {#sg_map_b} :vector<32x16xbf16> from vector<2x32x16xbf16> 
+  %vector_b = vector.extract_strided_slice  %vector_a { offsets = [0], sizes = [16], strides = [1]}: vector<16x2xbf16> from vector<32x2xbf16>
+
+  #sg_map_a = xegpu.sg_map<wi_layout = [1, 16], wi_data = [2, 1]> //the producer op of vector_a uses #sg_map_a
+  #sg_map_b = xegpu.sg_map<wi_layout = [1, 16], wi_data = [2, 1]>
+  //Before WI distribution:  %vector_b = vector.extract_strided_slice  %vector_a { offsets = [0], sizes = [32], strides = [1]}:  vector<32x16xbf16> from vector<64x16xbf16>
+  %vector_b = vector.extract_strided_slice  %vector_a { offsets = [0], sizes = [16], strides = [1]}: vector<16x2xf16> from vector<32x2xbf16>
+
+  #sg_map_a = xegpu.sg_map<wi_layout = [1, 16], wi_data = [1, 2]> //the producer op of vector_a uses #sg_map_a
+  #sg_map_b = xegpu.sg_map<wi_layout = [1, 16], wi_data = [1, 2]> 
+  #sg_map_c = xegpu.sg_map<wi_layout = [1, 16], wi_data = [1, 2]>
+  //Before WI distribution:  %vector_c = arith.addf  %vector_a, %vector_b {#sg_map_c} :vector<24x32xbf16>
+  %vec_a1, %vec_a2 = vector.deinterleave %vector_a : vector<24x2xbf16> to vector<24x1xbf16>
+  %vec_b1, %vec_b2 = vector.deinterleave %vector_b : vector<24x2xbf16> to vector<24x1xbf16>
+  %vec_c1 = vector.add  %vec_a1, %vec_b1 : vector<24x1xbf16>
+  %vec_c2 = vector.add  %vec_a2, %vec_b2 : vector<24x1xbf16>
+  %vector_c = vector.interleave %vec_c1, %vec_c2 :vector<24x1xbf16> to vector(24x2xbf16>
+```
+
+users must use for the WI data distribution of 1D block load and regular load with chunk_size on PVC and ARC. The rule of thumb is that the wi_data size must be 32bit, regardless on PVC or ARC.  Not using this sg_map defined here leads to undefined behavior.
 ```mlir
   For 1D block load
   # assert (wi_layout[0] x wi_layout[1] == subgroup_size) // PVC subgroup_size = 16
--- a/docs/rfcs/XeGPU.md
+++ b/docs/rfcs/XeGPU.md
@@ -328,9 +328,38 @@ Attribute `Memory_kind` describes the memory kind. "global" means the global mem
 
 ## XeGPU Attributes to support Work Item Level semantics
 
-Attribute `xegpu.sg_map` describes the mapping between work item (WI) and the 2D tensor specified by the tensor descriptor. To distribute the XeGPU operation to work items, the tensor_desc must be specified with the `sg_map` attribute at the tensor description creation time.
+**Attribute xegpu.sg_map**
+xegpu.sg_map specifies how a 2D tensor (defined by the tensor descriptor) is partitioned among work items (WIs) within a subgroup. sg_map consists of two parameters:
+  * wi_layout: Defines the 2D arrangement of WIs within the subgroup.
+  * wi_data: Specifies the shape of the tensor fragment that each WI loads or stores as a single packed data unit (16/32-bit).
 
-Within the sg_map, `wi_layout` specifies the layout of work items, describing the mapping of work items to the tensor. wi_layout[0] x wi_layout[1] must be equal to the subgroup size. `wi_data` specifies the minimum number of data elements assigned to each work item for a single distribution. In the example below, the subgroup has 16 work items in wi_layout=[1, 16], each accessing 1 element as specified by wi_data=[1,1].
+sg_map describes the pattern of a single packed data unit load/store done by all work items in a subgroup -- a distribution unit. Processing the entire tensor may require one or more distribution units. 
+
+When a sg_map attribute is attached to a tensor descriptor, load/store/dpas will operate in SIMT flavor. The sg_map attribute must be specified when creating the tensor descriptor.
+
+**Constraints**
+Given these definitions:
+```mlir
+wi_data_size = wi_data[0] × wi_data[1]
+subgroup_size = wi_layout[0] × wi_layout[1]
+distribute_unit_size = subgroup_size × wi_data_size
+tensor_size = tensor_desc[0] × tensor_desc[1]
+n_distribution_units = tensor_size / distribute_unit_size
+```
+
+the following conditions must hold:
+
+* subgroup_size must represent the number of WIs in a subgroup for a kernel.
+* tensor_size must be evenly divisible by distribute_unit_size (i.e., tensor_size % distribute_unit_size == 0).
+* tensor_desc[0] must be evenly divisible by wi_layout[0] × wi_data[0].
+* tensor_desc[1] must be evenly divisible by wi_layout[1] × wi_data[1].
+
+Note: When wi_data describes multiple elements, they must all come from a single, contiguous dimension.
+
+**Resulting WI Data Fragment**
+Each work item’s fragment of the distributed tensor is represented by a 2d vector (e.g., a SPIR-V or LLVM vector) with the shape [n_distribution_units, wi_data_size]. The result 2d vector will be further lowered to a 1d “SIMT-flavored” vector, such as a SPIR-V vector or LLVM vector, as the elements in the inner dimension being packed to a single packed data unit.
+
+In the example below, the subgroup has 16 work items in wi_layout=[1, 16], each accessing 1 element as specified by wi_data=[1,1]. So, wi_data_size is 1, distribute_unit_size is 16, tensor_size is 128.
 
 ```mlir
   #sg_map_a = xegpu.sg_map<wi_layout = [1, 16], wi_data = [1, 1]>
@@ -338,24 +367,6 @@ Within the sg_map, `wi_layout` specifies the layout of work items, describing th
 		: uint64, index, index, index, index, index, index
      	into tensor_desc<8x16xbf16, #sg_map_a>
 ```
-
-wi_data_size refers to the data size mapped to indiviudal work item, and distribute_unit_size represents the minimun size of 2D tensor to be distributed to all work items in a subgroup. tensor_size refers to the size of the tensor sepcified by tensor_desc. In the example above, wi_data_size is 1, distribute_unit_size is 16, tensor_size is 128.
-```mlir
-        wi_data_size = wi_data[0] x wi_data[1]
-	subgroup_size == wi_layout[0] x wi_layout[1]
- 	distribute_unit_size = subgroup_size x wi_data_size
-  	tensor_size = tensor_desc[0] x tensor_desc[1]
-```
-wi_data_size can be greater than 1, meaning each work item can own multiple elements in a single distribution. These multiple elements, indicated by wi_data, must come from one dimension and must be contiguous along that dimension.
-
-To distribute a tensor, tensor_size must be divisible by distribute_unit_size. Specifically, tensor_desc[0] must be divisible by wi_layout[0] x wi_data[0], and tensor_desc[1] must be divisible by wi_layout[1] x wi_data[1]. The 2D subtensor is distributed in a round-robin manner, meaning each work item receives multiple distributions. As a result, the distribution assigns a 2D data fragment to each work item, where the first dimension corresponds to the number of distributions, and the second corresponds to wi_data_size. The 2D data fragment is ultimately lowered to a "SIMT-flavored" vector, such as a SPIR-V vector or LLVM vector, where each vector element contains the packed data of the wi_data. The size of packed data could be 32-bit or 16-bit. 
-
-```mlir
-	WI_data_frag[0] = tensor_size / distribute_unit_size 
-	WI_data_frag[1] = wi_data_size 
-```
-
-The WI dsitribution is represented by the shape of the result vector being loaded, which is being reduced from [tensor_decs[0], tensor_desc[1]] to [WI_data_frag[0], WI_data_frag[1]].
 
 With `sg_map` attribute attached to tensor_desc, xegpu.load_nd operates in SIMT flavor and returns back a fragement associated with individual work item. The tensor_desc in the first example below specifies a tensor of 8x16 elements, which is distributed 8 times so each work item gets <8x1xbf16>. The second example shows the each work gets <8x2xint8> with 2 int8 elements packed as one 16-bit data.  
 ```mlir
@@ -366,6 +377,16 @@ With `sg_map` attribute attached to tensor_desc, xegpu.load_nd operates in SIMT 
   #sg_map_a = xegpu.sg_map<wi_layout = [1, 16], wi_data = [1, 2]>
   %vector_a = xegpu.load_nd %tdesc_a:
      tensor_desc<8x32xint8, #sg_map_a> into vector<8x2xint8>
+```
+The example below shows a larger 2d tensor being distributed using sg_map. 
+```mlir
+  #sg_map_a = xegpu.sg_map<wi_layout = [1, 16], wi_data = [1, 1]>
+  %vector_a = xegpu.load_nd %tdesc_a:
+     tensor_desc<12x32xbf16, #sg_map_a> into vector<24x1xbf16>
+
+  #sg_map_a = xegpu.sg_map<wi_layout = [1, 16], wi_data = [1, 2]>
+  %vector_a = xegpu.load_nd %tdesc_a:
+     tensor_desc<12x32xbf16, #sg_map_a> into vector<12x2xbf16>
 ```
 
 The example below shows the wi_data contains 2 elements for the first dimension. The result vector takes wi_data_size as inner dimension size, the data fragement <16x1xbf16> is loaded and packed as <8x2xbf16>, a process also known as "VNNI" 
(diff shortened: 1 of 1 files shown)


akroviakov:
I have reviewed the documentation update and wanted to take this opportunity to propose a clearer version (or at least a basis for the clearer version). When I first read the docs to understand sg_map and formalize it in terms of verification rules, I found the general description to be somewhat convoluted and vague. 

For example:

- The crucial term "distribution" (as in "receives multiple distributions") is never introduced.  
- "wi_data_size refers to the data size mapped to indiviudal work item" is ambiguous in terms of what "data size" is (elements, bytes), especially when we later mention data packing. 

I would like to propose a denser and slightly more formalized version (or a basis for it) of the sg_map description that will come before discussing concrete examples:
__________________________
**Attribute xegpu.sg_map**
`xegpu.sg_map` specifies how a 2D tensor (defined by the tensor descriptor) is partitioned among work items (WIs) within a subgroup. `sg_map` consists of two parameters:

- `wi_layout`: Defines the 2D arrangement of WIs within the subgroup.
- `wi_data`: Specifies the shape of the tensor fragment that each WI loads or stores as a single packed data unit (16/32-bit).

`sg_map` describes the pattern of a single packed data unit load/store done by all work items in a subgroup -- a distribution unit. Processing the entire tensor may require one or more distribution units.
When attached to a tensor descriptor, load/store/dpas will operate in SIMT flavor.

_Note_: The `sg_map` attribute must be specified when creating the tensor descriptor.

**Constraints**
Given these definitions:
```mlir
wi_data_size = wi_data[0] × wi_data[1]
subgroup_size = wi_layout[0] × wi_layout[1]
distribute_unit_size = subgroup_size × wi_data_size
tensor_size = tensor_desc[0] × tensor_desc[1]
n_distribution_units = tensor_size / distribute_unit_size
```
the following conditions must hold:
- `subgroup_size` must represent the number of WIs in a subgroup for a kernel.
- `tensor_size` must be evenly divisible by distribute_unit_size (i.e., `tensor_size % distribute_unit_size == 0`).
- `tensor_desc[0]` must be evenly divisible by `wi_layout[0] × wi_data[0]`.
- `tensor_desc[1]` must be evenly divisible by `wi_layout[1] × wi_data[1]`.

_Note_: When `wi_data` describes multiple elements, they must all come from a single, contiguous dimension.

**Resulting WI Data Fragment**
Each work item’s fragment of the distributed tensor is represented by a 2D “SIMT-flavored” vector (e.g., a SPIR-V or LLVM vector) with the shape `[n_distribution_units, wi_data_size]`.

Jianhui-Li:
> I have reviewed the documentation update and wanted to take this opportunity to propose a clearer version (or at least a basis for the clearer version). When I first read the docs to understand sg_map and formalize it in terms of verification rules, I found the general description to be somewhat convoluted and vague.

The proposed description looks great to me. Thank you Artem for the help. I am incorporating them.

akroviakov (approved):

docs/rfcs/XeGPU.md
```diff
@@ -333,9 +328,38 @@ Attribute `Memory_kind` describes the memory kind. "global" means the global mem
...
 
-Within the sg_map, `wi_layout` specifies the layout of work items, describing the mapping of work items to the tensor. wi_layout[0] x wi_layout[1] must be equal to the subgroup size. `wi_data` specifies the minimum number of data elements assigned to each work item for a single distribution. In the example below, the subgroup has 16 work items in wi_layout=[1, 16], each accessing 1 element as specified by wi_data=[1,1].
+When a sg_map attribute is attached to a tensor descriptor, load/store/dpas will operate in SIMT flavor. The sg_map attribute must be specified when creating the tensor descriptor.
+
+**Constraints**
+Given these definitions:
+```mlir
+wi_data_size = wi_data[0] × wi_data[1]
+subgroup_size = wi_layout[0] × wi_layout[1]
+distribute_unit_size = subgroup_size × wi_data_size
```
Garra1980, line 345: maybe distribut**ion**_unit_size?

docs/rfcs/XeGPU.md
```diff
@@ -344,52 +368,76 @@ Within the sg_map, `wi_layout` specifies the layout of work items, describing th
      	into tensor_desc<8x16xbf16, #sg_map_a>
 ```
 
-wi_data_size refers to the data size mapped to indiviudal work item, and sg_map_size to the collective size by all the work items as specified by sg_map. distribute_unit_size represents the minimun size of 2D tensor to be distributed to work items in a subgroup. tensor_size refers to the size of the tensor sepcified by tensor_desc.
-In the example above, wi_data_size is 1, sg_map_size is 16, tensor_size is 128.
+With `sg_map` attribute attached to tensor_desc, xegpu.load_nd operates in SIMT flavor and returns back a fragement associated with individual work item. The tensor_desc in the first example below specifies a tensor of 8x16 elements, which is distributed 8 times so each work item gets <8x1xbf16>. The second example shows the each work gets <8x2xint8> with 2 int8 elements packed as one 16-bit data.
```
Garra1980, line 371: The second example shows the each work gets <8x2xint8> with 2 int8 elements packed as one 16-bit data. ->
The second example shows the each work **item** gets <8x2xint8> with 2 int8 elements packed as one 16-bit data.

docs/rfcs/XeGPU.md
```diff
@@ -328,35 +328,46 @@ Attribute `Memory_kind` describes the memory kind. "global" means the global mem
...
 
-To distribute a tensor, tensor_size must be divisible by distribute_unit_size. Specifically, tensor_desc[0] must be divisible by wi_layout[0] x wi_data[0], and tensor_desc[1] must be divisible by wi_layout[1] x wi_data[1]. The 2D subtensor is distributed in a round-robin manner, meaning each work item receives multiple distributions. As a result, the distribution assigns a 2D data fragment to each work item, where the first dimension corresponds to the number of distributions, and the second corresponds to wi_data_size. The 2D data fragment is ultimately lowered to a "SIMT-flavored" vector, such as a SPIR-V vector or LLVM vector, where each vector element contains the packed data of the wi_data. The size of packed data could be 32-bit or 16-bit. 
+the following conditions must hold:
+
+* subgroup_size must represent the number of WIs in a subgroup for a kernel.
+* tensor_size must be evenly divisible by distribute_unit_size (i.e., tensor_size % distribute_unit_size == 0).
```
charithaintc, line 353: second condition is redundant and met by cond 3 and 4.

charithaintc, line 353: second condition is redundant and met by cond 3 and 4.

docs/rfcs/XeGPU.md
```diff
@@ -396,6 +417,26 @@ The examples below demonstrate how wi_data can be used to model the transpose_bi
...
+  #sg_map_a = xegpu.sg_map<wi_layout = [1, 16], wi_data = [1, 1]>
+  #tdesc_attr1 = !xegpu.block_tdesc_attr<memory_space=slm, boundary_check=false, sg= #sg_map_a>
+  %tdesc1 = xegpu.create_nd_tdesc %mem_addr, %offset :
+		uint64, index into tensor_desc<16xbf16, #tdesc_attr1>
+
+  %vector_a = xegpu.load_nd %tdesc_1:
+     tensor_desc<16xbf16, #sg_map_a> into vector<1x1xbf16>
+```	
+`xegpu.sg_map` also applies to 3d vector, which represents the result of 2d block load with array_length.  
+```mlir
```
charithaintc, line 430: maybe we can also consider removing 3d representation for array_length from XeGPU SG level as well for consistency.

charithaintc (approved):
LGTM

Jianhui-Li:
> I think we also need some clarification/section on the validity of XeGPU IR when sg_map is attached.
> 
> 1. WG level XeGPU/XeTile + sg_map -  is this valid IR?
> 2. SG level XeGPU + sg_map - this is only valid as an intermediate step during sg map propagation (not exposed to user?)
> 3. SIMT distributed XeGPU IR + sg_map - this the SIMT flavor of XeGPU and exposed to user.
> 
> @Jianhui-Li I think its better if we can also clarify these in the SIMT distribution section of the RFC.

All of these are valid IR. But some of passes we developed expect user to set wg_map and/or sg_map to be associated with all operations including vector/arith/math. These passes should be arranged as internal passes/phases, so that the high-level passes exposed to user don't assume wg_map/sg_map to associated with any non-XeGPU operations. 

If user just put the sg_map with the XeGPU.tensor_desc attributes in the XeGPU operations after WI distribution. It is valid and stable IR. so answer to question 3 is yes.

If user just put the sg_map with the XeGPU.tensor_desc attributes in the XeGPU operations before WI distribution, we expect user needs to set the sg_map for all non_xegpu operations and user our internal pass to perform WI distribution. So user needs to be responsible since these sg_map associated with non_xegpu ops is not stable and may be lost after applying other general up-stream passes.  so answer to question 2 is yes, I agree it is internal only. 

For the question 1, let's assume that XeGPU also adds wg_map.  I would expect user would like to own the blocking decision for wg distribution since it is performance critical, and then pass the decision using wg_map. Our pass may initially expect user to pass the wg_map to be associated with all non-xegpu ops, but the expectation should be removed evetually so even the wg_map associated with non-xegpu ops are lost, our wg distribution pass should work regardless. Unlike the sg_map attribute, the wg_map attribute should be gone after wg distribution, since we don't need it to validate the IR (to deal with inconsistent tensor desc shape and WI vector).

charithaintc:
> > I think we also need some clarification/section on the validity of XeGPU IR when sg_map is attached.
> > 
> > 1. WG level XeGPU/XeTile + sg_map -  is this valid IR?
> > 2. SG level XeGPU + sg_map - this is only valid as an intermediate step during sg map propagation (not exposed to user?)
> > 3. SIMT distributed XeGPU IR + sg_map - this the SIMT flavor of XeGPU and exposed to user.
> > 
> > @Jianhui-Li I think its better if we can also clarify these in the SIMT distribution section of the RFC.
> 
> All of these are valid IR - but really depends whether we have passes to process them. But some of passes we developed expect user to set wg_map and/or sg_map to be associated with all operations including vector/arith/math. These passes should be arranged as internal passes/phases, so that the high-level passes exposed to user don't assume wg_map/sg_map to associated with any non-XeGPU operations.
> 
> If user just put the sg_map with the XeGPU.tensor_desc attributes in the XeGPU operations after WI distribution. It is valid and stable IR. so answer to question 3 is yes.
> 
> If user just put the sg_map with the XeGPU.tensor_desc attributes in the XeGPU operations before WI distribution, we expect user needs to set the sg_map for all non_xegpu operations and user our internal pass to perform WI distribution. So user needs to be responsible since these sg_map associated with non_xegpu ops is not stable and may be lost after applying other general up-stream passes. so answer to question 2 is yes, I agree it is internal only.
> 
> For the question 1, let's assume that XeGPU also adds wg_map. I would expect user would like to own the blocking decision for wg distribution since it is performance critical, and then pass the decision using wg_map. Our pass may initially expect user to pass the wg_map to be associated with all non-xegpu ops, but the expectation should be removed evetually so even the wg_map associated with non-xegpu ops are lost, our wg distribution pass should work regardless. Unlike the sg_map attribute, the wg_map attribute should be gone after wg distribution, since we don't need it to validate the IR (to deal with inconsistent tensor desc shape and WI vector).

Thanks for the explanation. 

going from WG to SG is cleaner because we drop the wg_map after WG distribution. My concern is about SG level or SIMT XeGPU IR. It seems like we rely on "the presence of sg_map" in non-xegpu operations to determine if the IR is SG or SIMT. Also can you clarify why we still need sg_map after going to SIMT?

Jianhui-Li:
> It seems like we rely on "the presence of sg_map" in non-xegpu operations to determine if the IR is SG or SIMT. Also can you clarify why we still need sg_map after going to SIMT?

It is not TRUE we rely on "the presence of sg_map" in non-xegpu op to determine the IR is SG.  "the presence of sg_map" in non-xegpu op is an intermediate step inside the WI distribution pass.  A pass can also look at the size of result vector and see whether the inner-most dim match with wi_data_size. 

> Also can you clarify why we still need sg_map after going to SIMT?
It was mainly to keep the sg level information so that we can validate the IR independently. For load_nd, the tensor desc shape and WI vector shape will be different, so the validation will take sg_map to validate it.

docs/rfcs/XeGPU.md
Jianhui-Li: fixed.

docs/rfcs/XeGPU.md
Jianhui-Li: how?

charithaintc: I mean if the SIMT level IR does not show the array_len as an extra dimension then why should SG level show it?
```
tensor_desc<32x16, array_len = 2> -> vector<64x16>
reshape to 2x32x16
extract
```
instead of,
```
tensor_desc<32x16, array_len = 2> -> vector<2x32x16>
extract
```
This will be consistent with SIMT lowering logic.

Garra1980: @charithaintc How do you propose to change RFC?

charithaintc: I clarified this with Jian Hui already. no more changes are needed.

charithaintc:
@Jianhui-Li There is bit of a roadblock with the strategy. Consider this IR.
```
func.func @test_load_nd_distribution(%dst: !xegpu.tensor_desc<24x32xf16, #blk_tdesc, #sg_map_16>) -> (vector<48x1xf16>) {
  %laneid = gpu.lane_id
  %r = gpu.warp_execute_on_lane_0(%laneid)[16]
        args(%dst: !xegpu.tensor_desc<24x32xf16, #blk_tdesc, #sg_map_16>) -> (vector<48x1xf16>) {
    ^bb0(%arg0: !xegpu.tensor_desc<24x32xf16, #blk_tdesc, #sg_map_16>):
    %0 = xegpu.load_nd %arg0 <{l1_hint = #xegpu.cache_hint<cached>, l2_hint = #xegpu.cache_hint<uncached>}>
       : !xegpu.tensor_desc<24x32xf16, #blk_tdesc, #sg_map_16> -> vector<24x32xf16>
    gpu.yield%0 : vector<24x32xf16>
  }
  return %r : vector<48x1xf16>
}
```
This IR fails to verify in upstream because verification fails for the `gpu.warp_execute_on_lane_0`. reason is we are expanding dim 0 of the output vector due to round robin assignment. 24 becomes 48. 
```
error: 'gpu.warp_execute_on_lane_0' op expected expanded vector dimension #0 (24) to be a multipler of the distributed vector dimension (48)
  %r = gpu.warp_execute_on_lane_0(%laneid)[16]
```

Jianhui-Li:
> @Jianhui-Li There is bit of a roadblock with the strategy. Consider this IR.
> 
> ```
> #sg_map_16 = #xegpu.sg_map<wi_layout = [1, 16], wi_data = [1, 1]>
> #blk_tdesc = #xegpu.block_tdesc_attr<memory_space = global>
> func.func @test_load_nd_distribution(%dst: !xegpu.tensor_desc<24x32xf16, #blk_tdesc, #sg_map_16>) -> (vector<48x1xf16>) {
>   %laneid = gpu.lane_id
>   %r = gpu.warp_execute_on_lane_0(%laneid)[16]
>         args(%dst: !xegpu.tensor_desc<24x32xf16, #blk_tdesc, #sg_map_16>) -> (vector<48x1xf16>) {
>     ^bb0(%arg0: !xegpu.tensor_desc<24x32xf16, #blk_tdesc, #sg_map_16>):
>     %0 = xegpu.load_nd %arg0 <{l1_hint = #xegpu.cache_hint<cached>, l2_hint = #xegpu.cache_hint<uncached>}>
>        : !xegpu.tensor_desc<24x32xf16, #blk_tdesc, #sg_map_16> -> vector<24x32xf16>
>     gpu.yield%0 : vector<24x32xf16>
>   }
>   return %r : vector<48x1xf16>
> }
> ```
> 
> This IR fails to verify in upstream because verification fails for the `gpu.warp_execute_on_lane_0`. reason is we are expanding dim 0 of the output vector due to round robin assignment. 24 becomes 48.
> 
> ```
> error: 'gpu.warp_execute_on_lane_0' op expected expanded vector dimension #0 (24) to be a multipler of the distributed vector dimension (48)
>   %r = gpu.warp_execute_on_lane_0(%laneid)[16]
> ```

The alternative of vector<48x1xf16> is vector<24 x 1 x vector<1xf16>>. Does this nested vector work better? 
Or we just modify the upstream verification for the distribution rule.

charithaintc:
> > @Jianhui-Li There is bit of a roadblock with the strategy. Consider this IR.
> > ```
> > #sg_map_16 = #xegpu.sg_map<wi_layout = [1, 16], wi_data = [1, 1]>
> > #blk_tdesc = #xegpu.block_tdesc_attr<memory_space = global>
> > func.func @test_load_nd_distribution(%dst: !xegpu.tensor_desc<24x32xf16, #blk_tdesc, #sg_map_16>) -> (vector<48x1xf16>) {
> >   %laneid = gpu.lane_id
> >   %r = gpu.warp_execute_on_lane_0(%laneid)[16]
> >         args(%dst: !xegpu.tensor_desc<24x32xf16, #blk_tdesc, #sg_map_16>) -> (vector<48x1xf16>) {
> >     ^bb0(%arg0: !xegpu.tensor_desc<24x32xf16, #blk_tdesc, #sg_map_16>):
> >     %0 = xegpu.load_nd %arg0 <{l1_hint = #xegpu.cache_hint<cached>, l2_hint = #xegpu.cache_hint<uncached>}>
> >        : !xegpu.tensor_desc<24x32xf16, #blk_tdesc, #sg_map_16> -> vector<24x32xf16>
> >     gpu.yield%0 : vector<24x32xf16>
> >   }
> >   return %r : vector<48x1xf16>
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > This IR fails to verify in upstream because verification fails for the `gpu.warp_execute_on_lane_0`. reason is we are expanding dim 0 of the output vector due to round robin assignment. 24 becomes 48.
> > ```
> > error: 'gpu.warp_execute_on_lane_0' op expected expanded vector dimension #0 (24) to be a multipler of the distributed vector dimension (48)
> >   %r = gpu.warp_execute_on_lane_0(%laneid)[16]
> > ```
> 
> The alternative of vector<48x1xf16> is vector<24 x 1 x vector<1xf16>>. Does this nested vector work better?
> 
> Could be a pass to first use the nested vector to support the WI distribution as middle step, and then a separate pass to linearize from vector<24 x 1 x vector<1xf16>> to vector<48x1xf16> is that doable?
> 
> Or we just modify the upstream verification for the distribution rule.

How is vector<24 x 1 x vector<1xf16>> same as <48x1xf16>? is there a typo.

we can define a SIMT distributable unit as the inner vector and move those round robin dimensions out.
Example,
```
tensor_desc<24x32xf16, wi_data = [1, 1], wi_layout = [1, 16]> -> vector<24x2xvector<16xf16>>
```
gets lowered to
```
tensor_desc<24x32xf16, wi_data = [1, 1], wi_layout = [1, 16]> -> vector<24x2xvector<1xf16>>
```

charithaintc:
upstream `warp_execute_on_lane_0` has following restrictions.
1. ranks of source and distributed vectors must match.
2. each dim of distributed vector must be a multiple of source vector (we are violating this)
3. sourceDim[0]/distributedDim[0] * sourceDim[1]/distributedDim[1] * ... * sourceDim[rank-1]/distributedDim[rank-1] must be equal to warpSize

https://github.com/llvm/llvm-project/blob/b04a980b5597c61a8df2b489c4894bc0240b8e13/mlir/lib/Dialect/GPU/IR/GPUDialect.cpp#L2314

I think for now, only the following lowering is possible
```
tensor_desc<24x32xf16, wi_data = [1, 1], wi_layout = [1, 16]> -> vector<24x32xf16>
````
gets lowered to,
```
tensor_desc<24x32xf16, wi_data = [1, 1], wi_layout = [1, 16]> -> vector<24x2xf16>
```

I checked about vectors of vectors, but there are no examples currently in upstream mlir. So I am not sure it is a possible path. 

We could try to relax the constraints on  `warp_execute_on_lane_0` in upstream but there could be some push back because existing passes may depend on this constraint. this path needs to explored more.

**Comment by Jianhui-Li** [comment]
> upstream `warp_execute_on_lane_0` has following restrictions.
> 
> 1. ranks of source and distributed vectors must match.
> 2. each dim of distributed vector must be a multiple of source vector (we are violating this)
> 3. sourceDim[0]/distributedDim[0] * sourceDim[1]/distributedDim[1] * ... * sourceDim[rank-1]/distributedDim[rank-1] must be equal to warpSize
> 
> https://github.com/llvm/llvm-project/blob/b04a980b5597c61a8df2b489c4894bc0240b8e13/mlir/lib/Dialect/GPU/IR/GPUDialect.cpp#L2314
> 
> I think for now, only the following lowering is possible
> 
> ```
> tensor_desc<24x32xf16, wi_data = [1, 1], wi_layout = [1, 16]> -> vector<24x32xf16>
> ```
> 
> gets lowered to,
> 
> ```
> tensor_desc<24x32xf16, wi_data = [1, 1], wi_layout = [1, 16]> -> vector<24x2xf16>
> ```
> 
> I checked about vectors of vectors, but there are no examples currently in upstream mlir. So I am not sure it is a possible path.
> 
> We could try to relax the constraints on `warp_execute_on_lane_0` in upstream but there could be some push back because existing passes may depend on this constraint. this path needs to explored more.

My main point is that the WI distribution process is conceptually two-step process: The first step divides the 2D tensor data according to `wi_layout` to obtain a 2D subtensor. The second step extracts the elements to be packed from the dimension indicated by `wi_data`, treating it as the innermost dimension, and then linearizes the remaining elements in the 2D subtensor as the outermost dimension.

It seems that the current warp_execute_on_lane_0 mechanism is limited to solving the first step.  I was trying to propose  using vector of vector <24x2x<1xf16>> as the intermediate representation as the first step's output.  If it is not practical, we can just use vector<24x2xf16> (showing the subtensor value), which can be processed by the second step as long as the sg_map is provided. 

I updated the RFC to indicate the distribution is conceptually two step process.

**Comment by charithaintc** [comment]
> > upstream `warp_execute_on_lane_0` has following restrictions.
> > 
> > 1. ranks of source and distributed vectors must match.
> > 2. each dim of distributed vector must be a multiple of source vector (we are violating this)
> > 3. sourceDim[0]/distributedDim[0] * sourceDim[1]/distributedDim[1] * ... * sourceDim[rank-1]/distributedDim[rank-1] must be equal to warpSize
> > 
> > https://github.com/llvm/llvm-project/blob/b04a980b5597c61a8df2b489c4894bc0240b8e13/mlir/lib/Dialect/GPU/IR/GPUDialect.cpp#L2314
> > I think for now, only the following lowering is possible
> > ```
> > tensor_desc<24x32xf16, wi_data = [1, 1], wi_layout = [1, 16]> -> vector<24x32xf16>
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > gets lowered to,
> > ```
> > tensor_desc<24x32xf16, wi_data = [1, 1], wi_layout = [1, 16]> -> vector<24x2xf16>
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I checked about vectors of vectors, but there are no examples currently in upstream mlir. So I am not sure it is a possible path.
> > We could try to relax the constraints on `warp_execute_on_lane_0` in upstream but there could be some push back because existing passes may depend on this constraint. this path needs to explored more.
> 
> My main point is that the WI distribution process is conceptually two-step process: The first step divides the 2D tensor data according to `wi_layout` to obtain a 2D subtensor. The second step extracts the elements to be packed from the dimension indicated by `wi_data`, treating it as the innermost dimension, and then linearizes the remaining elements in the 2D subtensor as the outermost dimension.
> 
> It seems that the current warp_execute_on_lane_0 mechanism is limited to solving the first step. I was trying to propose using vector of vector <24x2x<1xf16>> as the intermediate representation as the first step's output. If it is not practical, we can just use vector<24x2xf16> (showing the subtensor value), which can be processed by the second step as long as the sg_map is provided.
> 
> I updated the RFC to indicate the distribution is conceptually two step process.

**one more clarification regarding this:** now this is a two stage process XeGPU op verification will be it tricky. Both the following become valid IR and should be able to pass the verifier.
```
after step 1 (naive distribution based on wi_layout) :
load_nd  .... : tensor_desc<24x32xf16, wi_alyout = [1, 16], wi_data = [1, 1]> -> vector<24x2xf16>
after step 2 (linearize/pack) :
load_nd  .... : tensor_desc<24x32xf16, wi_alyout = [1, 16], wi_data = [1, 1]> -> vector<48x1xf16>
```

**It would be nice if we could distinguish these two cases in the verifier by some way.** but currently we can not do it. both these versions must be allowed. 

I think it would be cleaner if at step 1 we only have `wi_layout` (no `wi_data`). Then before step 2, we do a packing propagation to introduce `wi_data` into the sg_map. what do you think?

**Comment by Jianhui-Li** [comment]
I think there is a way to put two stage process inside one pass, so that we only need to verify the final output.  Isn't so? 

>I think it would be cleaner if at step 1 we only have wi_layout (no wi_data). Then before step 2, we do a packing propagation to introduce wi_data into the sg_map. what do you think?

The propagation pass will need to propagate the sg_map as a whole, or else it make the design too complex.

**Comment by charithaintc** [comment]
> I think there is a way to put two stage process inside one pass, so that we only need to verify the final output. Isn't so?
yes, you can put two stages in one pass. But when you generate a new load op the verifier will be called automatically. I am checking if we can void calling the verifier. 

Otherwise, are we open to relaxing the checks on verifier?

**Review by chencha3 — APPROVED** [review]
LGTM. left some nit comments.

File: docs/rfcs/XeGPU.md [hunk 5]
```diff
@@ -333,9 +331,42 @@ Attribute `Memory_kind` describes the memory kind. "global" means the global mem
...
+n_distribution_units = tensor_size / distribution_unit_size
+```
+
+the following conditions must hold:
+
+* subgroup_size must represent the number of WIs in a subgroup for a kernel.
+* tensor_desc[0] must be evenly divisible by wi_layout[0] × wi_data[0].
+* tensor_desc[1] must be evenly divisible by wi_layout[1] × wi_data[1].
+
+As a result, tensor_size will be evenly divisible by distribution_unit_size (i.e., tensor_size % distribution_unit_size == 0), and each work item will recieve the distributed unit multiple times, with each unit having wi_data_size.
```
Garra1980, line 359: I guess distributed->distribution

docs/rfcs/XeGPU.md
```diff
@@ -344,52 +375,80 @@ Within the sg_map, `wi_layout` specifies the layout of work items, describing th
...
-
-The WI dsitribution is represented by the shape of the result vector being loaded, which is being reduced from [tensor_decs[0], tensor_desc[1]] to [WI_data_frag[0], WI_data_frag[1]].
-
-With `sg_map` attribute attached to tensor_desc, xegpu.load_nd operates in SIMT flavor and returns back a fragement associated with individual work item. The tensor_desc in the example below specifies a tensor of 8x16 elements, which is decomposed to 8x1 subtensors, each with sg_map_size 1x16. The result vector <8x16xbf16> is loaded and distributed to each WI as <8x1xbf16>.
+The example below shows a larger 2D tensor being distributed using sg_map.
 ```mlir
   #sg_map_a = xegpu.sg_map<wi_layout = [1, 16], wi_data = [1, 1]>
   %vector_a = xegpu.load_nd %tdesc_a:
-     tensor_desc<8x16xbf16, #sg_map_a> into vector<8x1xbf16>
+     tensor_desc<12x32xbf16, #sg_map_a> into vector<24x1xbf16>
```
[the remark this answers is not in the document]
Garra1980: > Why it is `vector<24x1xbf16>` not `vector<12x2xbf16>`?

I guess because of wi_data = [1,1]

docs/rfcs/XeGPU.md
```diff
@@ -344,52 +375,80 @@ Within the sg_map, `wi_layout` specifies the layout of work items, describing th
...
+#sg_map_a = xegpu.sg_map<wi_layout = [1, 16], wi_data = [1, 1]>
+%tdesc2 = xegpu.create_nd_tdesc %mem_addr, %offsets:2, %base_shape:2,%base_stride:2
+		: uint64, index, index, index, index, index, index
+     	into tensor_desc<8x16xbf16, #xegpu.block_tdesc_attr<array_length=2>, #sg_map_a>
+
+  %result = xegpu.load_nd %tdesc2 {L1_hint = uncached, L3_hint = uncached}:
+          tensor_desc<8x16xbf16, #xegpu.block_tdesc_attr<array_length=2>> into vector<16x1xbf16>
+```  
+
+xegpu.sg_map is also used to describe the WI data distribution for regular loads. The example below shows that each work item (WI) loads one fp32 data element. The resulting vector <16 x fp32> is loaded and distributed to each WI as <1 x fp32>.
```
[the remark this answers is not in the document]
Jianhui-Li: I am open to it.  

@charithaintc  any comments from implementation perspective?

docs/rfcs/XeGPU.md
```diff
@@ -344,52 +375,80 @@ Within the sg_map, `wi_layout` specifies the layout of work items, describing th
...
+```mlir
+  #sg_map_a = xegpu.sg_map<wi_layout = [1, 16], wi_data = [1, 2]>
+  #tdesc_attr1 = !xegpu.block_tdesc_attr<memory_space=slm, boundary_check=false, sg= #sg_map_a>
+  %tdesc1 = xegpu.create_nd_tdesc %mem_addr, %offset :
+		uint64, index into tensor_desc<16xbf16, #tdesc_attr1>
+
+  %vector_a = xegpu.load_nd %tdesc_1:
+     tensor_desc<16xbf16, #sg_map_a> into vector<1x2xbf16>
+```
+`xegpu.sg_map` also applies to 3D vector, which represents the result of 2D block load with array_length.
```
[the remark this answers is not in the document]
Jianhui-Li: It worth a separate discussion if arr_len is really long-standing hardware feature.

docs/rfcs/XeGPU.md
```diff
@@ -344,52 +375,80 @@ Within the sg_map, `wi_layout` specifies the layout of work items, describing th
...
-     tensor_desc<8x16xbf16, #sg_map_a> into vector<8x1xbf16>
+     tensor_desc<12x32xbf16, #sg_map_a> into vector<24x1xbf16>
+
+  #sg_map_a = xegpu.sg_map<wi_layout = [1, 16], wi_data = [1, 2]>
+  %vector_a = xegpu.load_nd %tdesc_a:
+     tensor_desc<12x32xbf16, #sg_map_a> into vector<12x2xbf16>
 ```
 
-For load_nd with `packed` attribute, wi_data[0] must equal to the size required to “pack” the low-precision data into 32-bit, also known as `vnni_factor`. The result vector takes wi_data[0] as inner dimension size, to indicate the effects of layout change known as "VNNI" transformation. The data fragement <16x1xbf16> is loaded and packed as <8x1x2xbf16>.
+The example below shows the wi_data contains 2 elements for the first dimension. The result vector takes wi_data_size as inner dimension size, the data fragement <16x1xbf16> is loaded and packed as <8x2xbf16>, a process also known as "VNNI" transformation.
```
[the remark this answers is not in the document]
Jianhui-Li: Right.

docs/rfcs/XeGPU.md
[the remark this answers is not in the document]
Jianhui-Li: We document the distribution rules in https://github.com/KhronosGroup/SPIRV-Registry/pull/306

N below means sg_size. 

**If N is equal to K, then each invocation contributes a single column of Matrix A. The complete contribution for an invocation is passed as a vector of M values, with each component representing one row of the contribution, and the lower-numbered components representing the lower-numbered rows.**

If N is less than K, then each invocation contributes multiple columns of Matrix A. The multiple columns are packed into a single value per row, with data from the lower-numbered columns in the lower bits of the value. The complete contribution for an invocation is passed as a vector of M values, with each component representing one row of the contribution, and the lower-numbered components representing the lower-numbered rows.

If N is greater than K, then each invocation contributes a fraction of the rows representing a single column of Matrix A. The complete contribution for an invocation is passed as a vector of values, with each component representing one row of the contribution, and the lower-numbered components representing the lower-numbered rows. For example, if N is twice as big as K, then each invocation contributes every other row of a single column of Matrix A, with the lower-numbered invocations contributing the first row and higher-numbered invocations contributing the second row. If the data from the higher-numbered invocations is not needed, such as when there is only one row M, then the data passed by the higher-numbered invocations is ignored.

docs/rfcs/XeGPU.md
```diff
@@ -344,52 +375,80 @@ Within the sg_map, `wi_layout` specifies the layout of work items, describing th
      	into tensor_desc<8x16xbf16, #sg_map_a>
 ```
 
-wi_data_size refers to the data size mapped to indiviudal work item, and sg_map_size to the collective size by all the work items as specified by sg_map. distribute_unit_size represents the minimun size of 2D tensor to be distributed to work items in a subgroup. tensor_size refers to the size of the tensor sepcified by tensor_desc.
-In the example above, wi_data_size is 1, sg_map_size is 16, tensor_size is 128.
+With `sg_map` attribute attached to tensor_desc, xegpu.load_nd operates in SIMT flavor and returns back a fragement associated with individual work item. The tensor_desc in the first example below specifies a tensor of 8x16 elements, which is distributed 8 times so each work item gets <8x1xbf16>. The second example shows the each work item gets <8x2xint8> with 2 int8 elements packed as one 16-bit data.
```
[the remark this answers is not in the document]
Jianhui-Li: The lower level ISA accepts 2 int8 elements packed as one unit.

docs/rfcs/XeGPU.md
Jianhui-Li: fixed

docs/rfcs/XeGPU.md
```diff
@@ -333,9 +331,42 @@ Attribute `Memory_kind` describes the memory kind. "global" means the global mem
...
+**Attribute xegpu.sg_map**
+xegpu.sg_map specifies how a 2D tensor (defined by the tensor descriptor) is partitioned among work items (WIs) within a subgroup. sg_map consists of two parameters:
+  * wi_layout: Defines the 2D arrangement of WIs within the subgroup.
+  * wi_data: Specifies the shape of the tensor fragment that each WI loads or stores as a single packed data unit (16/32-bit).
 
-Within the sg_map, `wi_layout` specifies the layout of work items, describing the mapping of work items to the tensor. wi_layout[0] x wi_layout[1] must be equal to the subgroup size. `wi_data` specifies the minimum number of data elements assigned to each work item for a single distribution. In the example below, the subgroup has 16 work items in wi_layout=[1, 16], each accessing 1 element as specified by wi_data=[1,1].
+sg_map describes the pattern of a single packed data unit load/store done by all work items in a subgroup -- a distribution unit. Processing the entire tensor may require one or more distribution units.
```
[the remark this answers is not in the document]
Jianhui-Li: rephrased it. thanks!

docs/rfcs/XeGPU.md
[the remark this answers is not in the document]
Jianhui-Li: may worth a separate discussion.

docs/rfcs/XeGPU.md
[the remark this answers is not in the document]
charithaintc: I am ok with it. 

however, what happens if we want to use it in DPAS for example? DPAS expect 2D layout.

docs/rfcs/XeGPU.md
```diff
@@ -412,10 +474,22 @@ Below example shows that each WI loads 4 fp32 data element with the chunk_size.
...
 
+The load with chunk_size pack the low-precision data to 32-bit data using wi_data = [1, 2].
+```mlir
+  #sg_map_t = xegpu.sg_map<wi_layout = [16, 1], wi_data = [1, 2]>
+  #scatter_attr = !xegpu.tdesc_attr< memory_space=slm, scattered=true>
+  %scatter_tdesc_chunk = xegpu.create_tdesc, %src_addr, %offsets
+		{chunk_size=4} :
+		uint64, vector<16xindex> into tensor_desc<16x8xbf16, #scatter_attr, #sg_map_t>
+
+  %result = xegpu.load_gather %scatter_tdesc_chunk, %mask {L1 = cached, L2 = uncached, transpose=[1,0]} :
```
Garra1980, line 485: btw I think in the upstream XeGPU def we have operation just named "load" instead of "load_gather"

Garra1980:
@Jianhui-Li @chencha3 @charithaintc anything else for array_length case or we're good to go with this PR?

charithaintc:
> @Jianhui-Li @chencha3 @charithaintc anything else for array_length case or we're good to go with this PR?

array_len discussion only applies to the SIMD representation AFAIK.

Jianhui-Li:
> > @Jianhui-Li @chencha3 @charithaintc anything else for array_length case or we're good to go with this PR?
> 
> array_len discussion only applies to the SIMD representation AFAIK.

I can change the PR to reflect our converged thoughts if we all agree on: 
1. using 2d vector as result of 2d load with array_len attribute
2. introducing 1d sg_map for WI distribution of 1d vector 

@chencha3  @charithaintc ?

charithaintc:
> > > @Jianhui-Li @chencha3 @charithaintc anything else for array_length case or we're good to go with this PR?
> > 
> > 
> > array_len discussion only applies to the SIMD representation AFAIK.
> 
> I can change the PR to reflect our converged thoughts if we all agree on:
> 
> 1. using 2d vector as result of 2d load with array_len attribute
> 2. introducing 1d sg_map for WI distribution of 1d vector
> 
> @chencha3 @charithaintc ?

I am ok with the change. 

for 1) there will be some implementation changes required in imex passes. For example optimize-transpose pass expect `vector.extract` after the array_len load. I am unsure about the impact on VNNI propagation. I guess it should be very minimal. 

For 2) we would need to tweak the upstream test cases again, I believe.

docs/rfcs/XeGPU.md
Jianhui-Li: OK. I will change the RFC to be consistent. 

https://mlir.llvm.org/docs/Dialects/XeGPU/#xegpuload-xegpuloadgatherop
xegpu.load (xegpu::LoadGatherOp)

Jianhui-Li: In a separate PR.

Jianhui-Li:
@Garra1980  I had a discussion with Chao and Charitha, we think the PR is ok to merge as current state. 

The following proposed changes are not essential so we would like to defer it for future to revisit. 
- using 2d vector as result of 2d load with array_len attribute
- introducing 1d sg_map for WI distribution of 1d vector

Garra1980 (approved):

PR #1131, opened 2025-11-13
Add a patch for profile support in upstream LevelZeroRuntimeWrapper.

mshahneo:
Add profile (execution time) support in upstream LevelZeroRuntimeWrapper.

We are using upstream LevelZeroRuntimeWrapper in IMEX SIMT/SIMD path. 
However, the upstream LevelZeroRuntimeWrapper does not have profiling support 
that we had in old downstream wrappers. This patch adds that support using 
LevelZero timstamp event-based profiling. One can use similar environment
variables used in downstream wrappers (i.e, `IMEX_ENABLE_PROFILING=1`)
to do different kind of profiling. 

This patch also fixes an issue described in internal issue: 1358.
The issue stems from maximum timestamp value calculation method.
Previously, the code made assumption that the deviceProperties.kernelTimestampValidBits
is always less than 64 bit, and hence use `(1ULL << deviceProperties.kernelTimestampValidBits) - 1ULL)`
to calculate maximum timeStampValue. It works fine if device_properties.kernelTimestampValidBits is 
less the size of ULL type in bits (usually 64 bits). However, it fails if if the number 
device_properties.kernelTimestampValidBits is >= ULL type in bits (64). This patch fixes that using:

```
zeTimestampMaxValue = (deviceProperties.kernelTimestampValidBits == 64)
                              ? std::numeric_limits<uint64_t>::max()
                              : ((1ULL << deviceProperties.kernelTimestampValidBits) - 1ULL);
    zeTimerResolution = deviceProperties.timerResolution;
```

### DIFF
--- a/build_tools/patches/0013-Add-profile-support-in-upstream-LevelZeroRuntimeWrap.patch
+++ b/build_tools/patches/0013-Add-profile-support-in-upstream-LevelZeroRuntimeWrap.patch
@@ -13,7 +13,7 @@ index d0728274b94c..c728e3576f11 100644
 +++ b/mlir/lib/ExecutionEngine/LevelZeroRuntimeWrappers.cpp
 @@ -13,12 +13,16 @@
  #include "llvm/ADT/Twine.h"
- 
+
  #include "level_zero/ze_api.h"
 +#include <algorithm>
 +#include <atomic>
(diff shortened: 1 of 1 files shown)


Garra1980:
Please fix pre-commit

PR #1231, opened 2026-08-26
[CI] Restrict CodeQL to the public repo and split off the C++ job

Garra1980:
Two follow-ups to #1230, now that the first scheduled/push runs have completed and we can see how it behaves.

### 1. Restrict the workflow to `intel/mlir-extensions`

Job is now guarded with `if: github.repository == 'intel/mlir-extensions'`, the same idiom `coverity.yml` already uses for its upload step.

### 2. Split the `c-cpp` analysis into its own scheduled-only job

The `push`/`pull_request` triggers are filtered to `'**/*.py'` and `'.github/workflows/**'`, but the matrix was not — so every run analysed all three languages, including the traced C++ build. On the first `main` run that took **1 h 58 m**, because the LLVM-MLIR cache written by `build.yml` had been evicted and LLVM was rebuilt from scratch. A one-line change to a Python script, or to this workflow file itself, would have been billed the same.

`analyze` now covers `actions` + `python` (no build, ~1 min) on push/PR/schedule, and `analyze-cpp` is gated to `schedule` and `workflow_dispatch`.

The split also lets each job state one build mode plainly, removing the four `if: matrix.build-mode == 'manual'` step conditions and the conditional `timeout-minutes: ${{ matrix.language == 'c-cpp' && 360 || 30 }}` expression.

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -23,10 +23,17 @@ concurrency:
   cancel-in-progress: ${{ github.event_name == 'pull_request' }}
 
 jobs:
+  # Interpreted languages and workflow files. These need no build, so they are cheap
+  # enough to run on every push and pull request.
   analyze:
     name: Analyze (${{ matrix.language }})
+    # Public repo only. sync.yml merges upstream/main into the innersource repo, so this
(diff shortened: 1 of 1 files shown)


mshahneo (approved):
LGTM.

PR #1240, opened 2026-09-04
[CI] Deduplicate the default LIT filter in XeGPU SIMT test script

Garra1980:
The default integration-test filter in `scripts/run_xegpu_simt_integration_tests.sh` was hardcoded in both cmake branches (LLVM-external-project build and prebuilt-LLVM build) and repeated a third time in the summary output, where it had already drifted: the printed directory list omitted `Integration/Dialect/Vector`, which the filter itself did include.

This change:
- hoists the directory list into a single `DEFAULT_LIT_FILTER_DIRS` array with a `default_lit_filter()` helper that joins it into the LIT `--filter` alternation;
- resolves `LIT_FILTER` once during option parsing, so both cmake branches use the same value;
- derives the printed "Test directories" list from that array, fixing the stale list;
- adds `--print-lit-filter`, which prints the effective filter regex (honoring `-t`) and exits, so CI workflows can query the pattern -- e.g. to build a complementary `--filter-out` for a `check-imex` run -- instead of maintaining a copy that can go out of sync;
- prints the actual regex in the final summary in both the custom-filter and default-filter cases.

No change to which tests are selected by default.

### DIFF
--- a/scripts/run_xegpu_simt_integration_tests.sh
+++ b/scripts/run_xegpu_simt_integration_tests.sh
@@ -33,6 +33,25 @@ print_section() {
     echo -e "${GREEN}========================================${NC}\n"
 }
 
+# Integration test directories owned by this script. This is the single source of
+# truth for the default LIT filter: it is used to build the --filter regex for the
+# cmake configurations below, and CI workflows query it via `--print-lit-filter` to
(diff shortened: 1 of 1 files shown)


mshahneo (approved):
LGTM.

